PR #3, opened 2019-09-18
Add dummy core library to train model; update config spec

constantinpape:


constantinpape:
@FynnBe 
I have added a small training script for the model that reads everything from the config to test that our design makes sense. I made some minor changes and it works now.
While doing this, I noticed a few more points we should address:
- transformations should support arguments as well
- currently, we don't support transformations on the labels and transformations that act jointly on the inputs and labels (necessary for augmentations)
- the idea with doi's for the data is nice, but after actually training a model, I can see a few limitations. For example, we want to be able to take data from some existing link without a doi. Maybe one can create a proxy doi without copying the data. (This is worth an issue on its own, but I wanted to bring it up already.)

constantinpape:
> * transformations should support arguments as well

Fixed by introducing `{name: ..., kwargs: {...}}` pattern for transforms as well.

> * currently, we don't support transformations on the labels and transformations that act jointly on the inputs and labels (necessary for augmentations)

For now, all transformations need to accept `x` and `y` (and should just return them unchanged if they don't act on it). This way, we can express all augmentations etc. with the entry points for transformations we have. 
 
> * the idea with doi's for the data is nice, but after actually training a model, I can see a few limitations. For example, we want to be able to take data from some existing link without a doi. Maybe one can create a proxy doi without copying the data. (This is worth an issue on its own, but I wanted to bring it up already.)

For now, we decided to have predefined datasets. This will be one of the major points to work on, I will make a separate issue about this later.

README.md
```diff
@@ -22,10 +22,11 @@ Must contain the keys:
 - `script`: script with model definition.
   - `source`: relative path from config file to the script
   - `hash`: hash of the script file
+  - `name`: Name of class defining the model in `script`
```
FynnBe: maybe keep `class_name`, to not confuse this key with the name of the script (file name).

README.md
```diff
@@ -90,23 +91,23 @@ training:
...
-            - Sigmoid
+        transformations:
+            - {name: Sigmoid}
     optimizer:
         name: Adam
-        kwargs: {beta: 0.01}
+        kwargs: {lr: 0.002}
     pretrained_on:
-        - {source: doi1, hash: 21}
+        - {name: NucleiDataset}  # optional: source : {identifier: doi/url, hash: hash_value}
```
FynnBe: maybe we can add the number of epochs trained here... This might be a useful metric, especially when training on several different datasets to get an idea of their influence.

README.md
```diff
@@ -22,10 +22,11 @@ Must contain the keys:
 - `script`: script with model definition.
```
FynnBe: key is not up to date with the the example in `example_configuration.yaml`

README.md
constantinpape: Hmm, I preferred the symmetry with the other options, now it's `name` and `kwargs` everywhere.

README.md
constantinpape: Good point.

README.md
FynnBe: yeah, nice solution with `definition`

README.md
```diff
@@ -47,9 +47,11 @@ Must contain:
...
   - `name`: Name of the optimizer (must)
   - `kwargs`: keyword arguments for the optimizer (can)
-- `pretrained_on`: list of datasets the model was trained on, valid entry:
-  - `name`: Name of dataset in the core library.
-  - TODO need optional ways of specifying identifiers for custom training data via url / doi
+- `pretrained_on`: description of pre-training:
+  - `datasets`: list of datasets the model was trained on, valid entry:
+    - `name`: Name of dataset in the core library.
+    - TODO need optional ways of specifying identifiers for custom training data via url / doi
+  - `n_epocs`: Number of epochs used in training (change to iterations ?)
```
FynnBe: yes, actually `n_iterations` could be more accurate and is independent of dataset size (slightly biased by batch size though).
if we are keeping `n_epoch`s anyway, I'd suggest to not abbreviate away the `h`

constantinpape: yeah, I will change to `n_iterations` later.
I was just to lazy to implement it in training, but I think it is better, because it says much more then the iteration count.

constantinpape:
@FynnBe I think we have addressed all initial points now and this is a good basis for further discussion.
I went ahead and merged it.

oeway:
Nice! You guys are moving fast!

constantinpape:
> Nice! You guys are moving fast!

Yes, excited to see this happen :).
We will discuss a bit more internally and let you know once we think it's a good basis for discussion.
Then we can start to flesh out issues / alternatives etc.

PR #6, opened 2019-10-30
Add repository manifest and schema file

oeway:
This PR adds a repository manifest file for all the models and transforms, it should:
 - named exactly `bioimage_manifest.yaml`
 - placed under the root directory of the Github repo

An empty schema file is also added which we need to fill up later.

### DIFF
--- a/bioimage_manifest.yaml
+++ b/bioimage_manifest.yaml
@@ -1,0 +1,8 @@
+name: Example Models
+description: A repository for example models
+
+models:
+  - UNet2d.model.yaml
+
+transforms:
+  - Normalization.transform.yaml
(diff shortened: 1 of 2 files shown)


bioimage_manifest.yaml
```diff
@@ -0,0 +1,8 @@
+name: Example Models
+description: A repository for example models
+
+models:
+  - UNet2d.model.yaml
+
+transforms:
+  - Normalization.transform.yaml
```
FynnBe: currently in use is the somewhat bulky `NormalizeZeroMeanUnitVariance` transformation...

bioimage_manifest.yaml
```diff
@@ -0,0 +1,8 @@
+name: Example Models
+description: A repository for example models
+
+models:
+  - UNet2d.model.yaml
+
+transforms:
```
FynnBe: we agreed on the use of `transformation` over `transform`, therefore I'd suggest to use `transformations` here.

PR #18, opened 2020-02-24
Add `license` field and rename `thumbnail` to `covers`

oeway:
This PR add license to the to the configuration and rename `thumbnail` to `covers`.

The point of changing `thumbnail` to `covers` is that we may need to display multiple images for the same model to better represent what the model does. You can imagine a gallery of example input/output images for the model. Therefore I would prefer `covers`or `thumbnails` than `thumbnail`. I am also thinking perhaps we can have both types, say a `thumbnail` image represent the model itself, and a `gallery` which contains a list of example input/images. Please comment below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ Relative path to file with additional documentation in markdown.
 A list of tags.
 
 ### `covers`
-A list of cover images provided by either a relative path to the model folder, or a hyperlink starts with `https`. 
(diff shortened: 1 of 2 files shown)


oeway:
@FynnBe do you have an image for the UNet2dExample.model.yaml? I think it would be great if we can add one here.

I propose we use an image of aspect ration 2:1, so we can display input/output in a left-right layout.

PR #25, opened 2020-06-18
Small changes to weight definition

constantinpape:


constantinpape:
@oeway we just discussed that we would like to organise the weights a bit differently:
- have the weights as a list instead of dict; the field `name` will correspond to the dict key
- add the fields `parent` and `description`

I added a second weight field now to the example for educational purposes, but they still refer to the same set of weights; we can also remove this again.

@m-novikov @FynnBe, could you check that this corresponds to the changes you had in mind?

models/UNet2dExample.model.yaml
```diff
@@ -25,10 +25,18 @@ model:
...
+      source: 10.5281/zenodo.3446812
+      sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
+      parent: null
+      description: "Weights from DataScienceBowl"
+      # These should be different they are not
+    - name: my_other_weights
+      source: 10.5281/zenodo.3446812
+      sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
+      parent: null
+      description: "Weights from DataScienceBowl"
```
FynnBe, line 39: Looking at it again I see that it's different from what I actually had in mind to represent an ordered dict:
```suggestion
    -  default:
        # TODO how do we make clear that this is a doi? doi:.... ?
        source: 10.5281/zenodo.3446812
        sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
        parent: null
        description: "Weights from DataScienceBowl"
      # These should be different they are not
    - my_other_weights:
        source: 10.5281/zenodo.3446812
        sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
        parent: null
        description: "Weights from DataScienceBowl"
```
I would not mind this format here either. If we go with this I would suggest (like @m-novikov mentioned in our brief morning discussion) to differentiate between `id` and `name`
```suggestion
    - id: default
      name: "Default weights"
      # TODO how do we make clear that this is a doi? doi:.... ?
      source: 10.5281/zenodo.3446812
      sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
      parent: null
      description: "Weights from DataScienceBowl"
      # These should be different they are not
    - id: my_other_weights
      name: "My other Weights"
      source: 10.5281/zenodo.3446812
      sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
      parent: null
      description: "Weights from DataScienceBowl"
```

Maybe adding `name` to the first 'ordered dict' option would avoid confusion, as then there is the ordered dict `key` and a human friendly `name`, just like `description`, etc...

oeway: For other implementation like in Javascript, it's more natural to wrap it as a regular nested object/dictionary. OrderedDict is not a standard type that we can easily express with JSON schema for example,  and the order of the weights should not matter.

So I would vote for the first suggestion by @FynnBe  above.

oeway: > @oeway we just discussed that we would like to organise the weights a bit differently:
> 
> * have the weights as a list instead of dict; the field `name` will correspond to the dict key

Why do we want a list instead of a dict? My concern is that if we wrap the name/id in a list, the uniqueness of the name/id is not guaranteed, and this cannot be expressed in JSON schema. If you think it's important to express the order, would a orderedDict work?

I would keep the dict, and simply add the name into the weights dict, and give the `name` a different definition -- a human friendly name. See my comments above.

> * add the fields `parent` and `description`
Sounds good

m-novikov: > Why do we want a list instead of a dict?
If we want to store checkpoints for example, they have defined order in which they should be displayed in client app user interface.

> My concern is that if we wrap the name/id in a list, the uniqueness of the name/id is not guaranteed, and this cannot be expressed in JSON schema.
We can't enforce all rules using only JSON schema, there will be always some kind business logic, so for me it seems okay to use validation code for these cases.

constantinpape: Looking at this again, I would agree with @oeway and leave it as a dict.
@m-novikov we can use the parent field to express the order.

FynnBe: I agree with @m-novikov ,  we have a whole lot of additional checks to perform anyway. Adding that each dictionary in the list has length one is simple enough I'd say...
```yaml
weights:
   - default:
        name: Default Model Weights
        source: 10.5281/zenodo.3446812
        sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
        parent: null
        description: This is an amazingly accurate description of these default model weights.
```
Alternatively we need a clear way of ordering the unordered dict. A `timestamp` field could be used (might be useful regardless if used for sorting or not).

constantinpape: Ok, I don't really mind either way; but I still see @oeway's point that the OrderedDict approach is not really a natural fit for other languages.

oeway: Right, I would suggest to use a plain dictionary instead of an ordered one.

Also, we should perhaps try to get rid of the orderedDict in other places in the spec, to make it much easier to write the rules into a json schema, and also for implementing it in other languages. Even in Python, the `OrderedDict` type is only available if we import it. In JS, there is a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) type, but it's rare to see the use of it.

constantinpape: > How do we represent order of checkpoints in this case?

I think the parent is the solution for this, for checkpoints we can easily include this.
And we could also include the `timestamp` that @FynnBe suggested, but I don't think it's strictly necessary.

> I would prefer not to use ordered dict, but list of dicts, which is available in any language.

What's not so nice about lists is that we have to refer to the weights than by index rather than by key,

m-novikov: > I think the parent is the solution for this, for checkpoints we can easily include this.
And we could also include the timestamp that @FynnBe suggested, but I don't think it's strictly necessary.

But this solution is more complicated it expects us to construct linked list given configuration.

> What's not so nice about lists is that we have to refer to the weights than by index rather than by key,

But dictionaries can contain and id or key which can be used for reference and uniqueness would be validated by parser.

constantinpape: > But dictionaries can contain and id or key which can be used for reference and uniqueness would be validated by parser.

You mean lists, right?
Yes, that could also be done.

FynnBe: if we have it unordered I would really add a timestamp. In doubt which weights to take one would usually like to take the newer ones... and we would get a free meaningful order for display purposes.

oeway: Yes, I think we should add timestamp.

constantinpape:
Ok, I think we have two competing solutions now:

Using a dict:
```
weights1:
  name: Some name
  description: Some description
  source: weights_source_uri
  sha256: adasdasf
  parent: null
  timestamp: time
weigths2:
  ...
```
Using a list:
```
- id: weights1
  name: Some name
  description: Some description
  source: weights_source_uri
  sha256: sasdaasda
  parent: null
  timestamp: time
-  id weigths2
   ... 
```
Note, we would support reference by `id` in the list solution.

FynnBe:
One more argument for the list:
if we have 'groups of weights', let's say w1_mouse, w2_mouse, w1_fly, w2_fly their order in time might not be the most intuitive order and let's say both 'w1' do not have a parent weight it cannot be ordered through the parent either. However a list allows for a manual order:
```yaml
- id: w1_mouse
  timestamp: old
  ...
- id: w2_mouse
  timestamp: new
  ...
- id: w1_fly
  timestamp: old
  ...
- id: w2_fly
  timestamp: new
  ... 
```

oeway:
No strong opinion here, I think there are pros and cons for both. I have implemented both formats in the new version of https://bioimage.io/ 

See a demo here: https://bioimage.io/#/?repo=%2Fstatic%2Ftest_manifest.bioimage.io.json&id=2sUNetDA
Scroll down in the dialog, and you can see two weights, note you can click on the table title to sort the weights by name or by timestamp. This means even if we use the dict format, one can easily sort by name for 'groups of weights'.

How about we go for the list with id solution and get this PR merged? It is currently blocking other tasks.

FynnBe:
Yes, let's move forward with these changes for now.

Your mentioning of `datasets` got me thinking that this might get somewhat out of hand, as weights don't arise from datasets but instead from a `training`... I'll make another issue to discuss this further.

oeway:
I updated the docs to reflect what we discussed above, will merge now, please send another PR if you want to change it further.

PR #26, opened 2020-06-18
Bump format version to 0.2.0

oeway:
We forgot to bump the format version when changing the specification. This is a patch PR for that, I bumped the version to `0.2.0`.

However, following Semantic Versioning 2.0.0 https://semver.org/ it seems we should bump the version to 1.0.0 since we are breaking the API, but I hesitate to do that because we are not ready to release it as 1.0.0 yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The configurations are represented by a yaml file.
 
 To get a quick overview of the config file, see an example file [here](./models/UNet2dExample.model.yaml).
 
-## Current `format_version`: 0.1.0
+## Current `format_version`: 0.2.0
 
 ## Common keys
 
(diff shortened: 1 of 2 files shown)


MartinHjelmare:
As long as we're not beyond 1.0 I think it's ok to make breaking changes without bumping major version.

constantinpape:
Yes, breaking changes in the Minor version are fine as long as Major is still 0.

constantinpape:
@oeway while I approve of bumping the version to 0.2.0, should we maybe wait till we resolve and merge #25?

oeway:
> @oeway while I approve of bumping the version to 0.2.0, should we maybe wait till we resolve and merge #25?

Yes, let's wait for that.

PR #30, opened 2020-10-15
[WIP] slimmed specification format 0.3.0 

FynnBe:
key differences:
- Specify framework per weight (to ease exporting models to other formats while keeping the source reference)
- absorb pre- and postprocessing into the model object to ease sharing across consumers

example use case: a trained model with pytorch source code has pytorch weights (which require the use of the pytorch source code), some weights were converted to tensorflow.pb or ONNX files, which can be used without the pytorch source code!

pytorch-bioimage-io PR: https://github.com/bioimage-io/pytorch-bioimage-io/pull/21


README.md
```diff
@@ -1,224 +1,184 @@
...
 Relative file path to test input. `language` and the file extension define its memory representation.
 The input is always stored as list of tensors.
 
-### `test_output`
+- `test_output`
 Relative file path to test output. `language` and the file extension define its memory representation.
 The input is always stored as list of tensors.
 
-### `inputs`
+- `inputs` <!-- Force this to be explicit, or also allow any? -->
```
oeway, line 109: I would prefer explicit

oeway:
Looks good. 

I am thinking whether we should make it even simpler to also take out the readers and samplers, because that's for training, one possibility is to move them to vendor specific config for now. Would be nice to make a minimal working spec for the initial release. What do you think?

FynnBe:
> I am thinking whether we should make it even simpler to also take out the readers and samplers

Sampler definitely 👍 Readers I'd aggree too for now. maybe we should call them datasets and have them as one for all languages/frameworks...  to there is no double entry, etc... 
We could put that decision off to later and just leave them out for now.

m-novikov:
Question about multiple weights, the initial idea was about storing checkpoints for the same model. 

The proposed change is supposed to store weights for different frameworks but solving the same task using the same model architecture.
The purpose is to enable grouping on the website. On the last meeting (16th October) we discussed that there are multiple possible grouping methods (user task, dataset, model architecture). 
Why we are treating model architecture differently from any other possible grouping/filtering?

What is the expected workflow if a developer wants to work on the model (e.g. adding his own weights and then deciding he wants to train using a different cost function/preprocessing)?

oeway:
@m-novikov I think for the website, we can use the tags field of each weights to achieve grouping and filtering.

For the model architecture, now I guess called `format`, is mainly used for facilitate the parsing/running of the models, e.g. with the model runners.

For training, you need to look into the source code of the model, and we do not guarantee that all the weights can be used for training, e.g. if you convert into ONNX format.


@FynnBe I remembered that you moved the `documentation` key from the root node to `model`, I would suggest to keep it in the root, so it doesn't introduce extra complexity to the parsing on the website. Basically we treat the model yaml as a RDF with all the basic keys, and now we just need to support treating also each weights item as a RDF.

m-novikov:
> I think for the website, we can use the tags field of each weights to achieve grouping and filtering.

The question is why we use a model file for one search dimension and tags for others.

My concern is if the model file is atomic then it's much easier to put implementation around it, but with more dimensions (e.g. framework variability, multiple weights) the runner implementation has more edge cases there are and UI becomes non trivial. (e.g. now I can't say checkpoins widget only for checkpoins, now I need to be aware that loading weights can change framework)

oeway:
> My concern is if the model file is atomic then it's much easier to put implementation around it, but with more dimensions (e.g. framework variability, multiple weights) the runner implementation has more edge cases there are and UI becomes non trivial. (e.g. now I can't say checkpoins widget only for checkpoins, now I need to be aware that loading weights can change framework)

I understand the complexity, but I still think these cannot be solved by making the model file atomic. Because otherwise we will just split the model file into several files and we need to handle the framework and weight format correspondingly. If you like, from the website, we can let the user export a model file that contains only one framework and one weights item. I doubt there will be a big difference in terms of implementation.

m-novikov:
To expand on my question:

> What is the expected workflow if a developer wants to work on the model (e.g. adding his own weights and then deciding he wants to train using a different cost function/preprocessing)?

Let's say someone develops a model `MyAwesomeUnet` then put the model and puts it on github.
Now, this person also responsible for accepting pull-requests for this model.
Would he want to accept the pull-request with converted models or retrained variations?
From the maintainer perspective, it seems questionable to accept changes with additional weights. Also what happens if person want's to "archive" their repo when for example finishing their PhD.

FynnBe:
> Would he want to accept the pull-request with converted models or retrained variations?

Yes, if the model source stayed the same. As we now absorb preprocessing into the model, if one would change the preprocessing for inference, the model source would have to change...

> From the maintainer perspective, it seems questionable to accept changes with additional weights. 

Could you elaborate, why?

> Also what happens if person want's to "archive" their repo when for example finishing their PhD.

Maybe in that case we could fork the corresponding repo to bioimage orga?

m-novikov:
> Could you elaborate, why?

Because while the original model was tested by the author, while pull-request uses a foreign framework that requires maintenance and as a result knowledge of how it was done which is may not be available to the author.

FynnBe:
Could it not be tested with the appropriate model runner? Ideally as CI of course... But we could provide an online tool, e.g. a BioEngine link..?

oeway:
I think this is a good point. And fork the model file will result in two duplicated model with the same name but different weighs. But in that case the website can taken care of composing weights from different sources.

For example, if we already have the model file, user can submit individual weights file contains only the weights fields and a model ID. Like this we don't need to have a physical single file with all the weights, but the website can dynamically aggregate model weights by id, and generate the composed model file for users.

FynnBe:
While comparing the current status to https://github.com/bioimage-io/bioimage.io/blob/master/docs/resource-description-file.md I see some discrepancies we need to resolve:

- [ ] while we model specification requires documentation to be a local .md file, bioimage.io rdf expects a url
- [ ] similar case: covers. Here we have covers under the model key as relative local files, and covers in root as urls in the rdf... this is confusing.

FynnBe:
> Like this we don't need to have a physical single file with all the weights, but the website can dynamically aggregate model weights by id, and generate the composed model file for users.

Thinking along these lines and @m-novikov 's suggestion for an atomic model config, I'm thinking we should add a weight's specification and make light weight weight configurations. The model architecture config would not need a single weight in it...

oeway:
> > Like this we don't need to have a physical single file with all the weights, but the website can dynamically aggregate model weights by id, and generate the composed model file for users.
> 
> Thinking along these lines and @m-novikov 's suggestion for an atomic model config, I'm thinking we should add a weight's specification and make light weight weight configurations. The model architecture config would not need a single weight in it...

I would suggest keeping the current format definition and treat weight file as a subset of the model file (e.g. add a `parent` key to specify that this weight should be attached to a model file). In other words, a weight file is a model file, and the website will combine all of them during the manifest compilation step.

As I mentioned in a previous message, I don't see the advantage of having an atomic model config, eventually we will need to implement all the weight format options, either atomic or not. Besides that a model consumer can choose to treat the current model file as atomic if it simply take the first weights definition. This would work easily if we combine with the export feature on the website. Say, if you decided that you only want to single weights in a file, we can ask user to select a weight and export as an atomic model config on the website.

I afraid separating a single model file into model file + weights files would further complicated the spec, and make it even more complicated to handle by different model consumers and decrease the portability of the model file. For example,  to actually run a model you will need at least two files (think how to download two files from the website, we are likely need to zip them).

My suggestion is to move forward with the current format, and let the website handle the complication of combining weights to the model file, for the model consumer, it will get a single yaml file with model and the weights. And perhaps for the sake of time, we can just skip this feature for now and assume don't need to combine weights files into model file.

m-novikov:
> Thinking along these lines and @m-novikov 's suggestion for an atomic model config, I'm thinking we should add a weight's specification and make light weight weight configurations. The model architecture config would not need a single weight in it...

There is no benefit in separating model file and weights.
A list of weights is quite useful for storing checkpoints of the same model of the same training process for the same data.
Enabling external grouping also allows us to have flexibility in what we consider the "same" model consistently across the whole bioimage.io website. Even if definition of the "same" model would change in the future it would be easy to adapt if it is not hardcoded in every model.

FynnBe:
> * [ ] while we model specification requires documentation to be a local .md file, bioimage.io rdf expects a url
> * [ ]  similar case: covers. Here we have covers under the model key as relative local files, and covers in root as urls in the rdf... this is confusing.

so, covers and documentation just needs to be a valid URI, and we put them in root?

esgomezm:
Hi all, 

here are some comments regarding the new changes:
> * absorb pre- and postprocessing into the model object to ease sharing across consumers

Shouldn't this one be (maybe) software/reader-specific? At the moment, in the example (https://github.com/bioimage-io/configuration/blob/slim_spec/models/UNet2dExample.model.yaml) it is at config/ilastik_training. 

Also, one could think, especially for the post-processing, as "weight-specific". For example, if there is an optimal threshold calculated to binarize the output of the model and this binarization is part of the post-processing. I'm not sure which is the best way to do it but it seems clear to be at least reader/software specific (Ilastik, Fiji, ZeroCostDL4Mic, Imjoy or DeepImageJ)

Regarding some specific fields in the documentation: 

* `covers`: For simplicity of the config.yaml, it might make sense to have covers for different `weights` as those produce different results (similar to the test input/output), but a cover in the `model` specifications may introduce some "noise" in the web-page appearance.

* `dependencies` in `model`: could be misled with the dependencies for the pre and post-processing steps. I'm not sure if we can accept those here...

To consider changing the example (I can make a PR):

* `framework` and `language`: in the example now they are at both places the root and the `model` specifications.

FynnBe:
> Shouldn't this one be (maybe) software/reader-specific? At the moment, in the example (https://github.com/bioimage-io/configuration/blob/slim_spec/models/UNet2dExample.model.yaml) it is at config/ilastik_training.

I just parked the incomplete (old) ilastik training spec at config/ilastik_training. Please ignore this for now. 
The idea to absorb pre-and postprocessing in the model results in a change of the model source file.
To alleviate the transition @m-novikov and I propose to add `normalization` to each entry in `inputs`. For now we would only support `null` or `zero_mean_unit_variance`. With this information provided the consumer can normalize the data beforehand.
As far as we see only this normalization was used in the preprocessing of the published models. Therefore, by adding the normalization key, we can leave the old model source codes untouched, add the appropriate normalization scheme and have format 0.3.0 ready quickly.



> Also, one could think, especially for the post-processing, as "weight-specific". For example, if there is an optimal threshold calculated to binarize the output of the model and this binarization is part of the post-processing. I'm not sure which is the best way to do it but it seems clear to be at least reader/software specific (Ilastik, Fiji, ZeroCostDL4Mic, Imjoy or DeepImageJ)

This sort of post-processing I would move into the model source. If global data for the transformation is needed it would be required as an additional input to the model, e.g.:
```yaml
inputs:
   - name: raw
     .....
   - name: threshold for binarization
     ....
```

FynnBe:
> * `covers`: For simplicity of the config.yaml, it might make sense to have covers for different `weights` as those produce different results (similar to the test input/output), but a cover in the `model` specifications may introduce some "noise" in the web-page appearance.

I think the idea with the covers is that the `covers` in root are for visualizing the architecture, etc. whereas the `weigh:covers` show the use case, as you pointed out. How we display them systematically without introducing too much noise is an issue of the website or d you think we should remove `root:covers` categorically?

> * `dependencies` in `model`: could be misled with the dependencies for the pre and post-processing steps. I'm not sure if we can accept those here...

when pre- and postprocessing is removed as planned, this should be clear.


> To consider changing the example (I can make a PR):
> 
> * `framework` and `language`: in the example now they are at both places the root and the `model` specifications.

Yes, this is not correct yet, happy to review a PR 👍

esgomezm:
> As far as we see only this normalization was used in the preprocessing of the published models.

I'm not completely sure. N2V for example needs some parameters extracted from the training for the preprocessing. One of our models does also recover the pixel size to rescale the image before the inference... that's why I suggested keeping them specific to software, maybe in the `config` field. 

> Therefore, by adding the normalization key, we can leave the old model source codes untouched, add the appropriate normalization scheme and have format 0.3.0 ready quickly.

> This sort of post-processing I would move into the model source. If global data for the transformation is needed it would be required as an additional input to the model, e.g.:

Sorry, I'm not sure whether I understand. You propose to add keys for standard transformations that do not require input parameter (standard normalization) and, if parameters are needed (percentile normalization, upsampling, changing pixel size), then, add them as one more input (an output for the post-processing case)? I see this a bit complicated as it would imply defining pre/post-processings in two different ways, isn't it?

I can only talk for DeepImageJ case: both processings are static IJ macros written for a specific task, then you would suggest adding some key that DeepImageJ understands and the macros as attachments?

FynnBe:
> I'm not completely sure. N2V for example needs some parameters extracted from the training for the preprocessing. One of our models does also recover the pixel size to rescale the image before the inference... that's why I suggested keeping them specific to software, maybe in the `config` field.

Sorry to have overlooked that use case. Keeping it consumer specific in `config/<consumer>` is always an option 👍 You could potentially paste a whole macro script in there.

> Sorry, I'm not sure whether I understand. You propose to add keys for standard transformations that do not require input parameter (standard normalization) and, if parameters are needed (percentile normalization, upsampling, changing pixel size), then, add them as one more input (an output for the post-processing case)? I see this a bit complicated as it would imply defining pre/post-processings in two different ways, isn't it?

It would be interesting to add more normalization schemes and expectation about the pixel size to the input specification. @m-novikov and I thought it would be great to have some defined values for `normalization` (and maybe as well a scheme for `physical_pixel_size`?) that describe the expected input more specifically. Transforming the input accordingly would be left to the consumer. Anything more complicated (or not aggreed upon yet in the spec format) need to happen in model source code (or consumer specific sections in `config`. 

I agree with you that this split is not ideal, but this would allow us to slowly extend our common ground as consumers, while keeping the flexibility we need to describe any model we might need. I suppose for IJ there is a clear split of tensorflow model and macro script pre-/postprocessing. In case of keras/tensorflow/pytorch the preprocessing that is moved to model source code can (in many cases) directly be serialized as part of the model. E.g. if I write a tensorflow model that includes its preprocessing, you can directly run it in DeepImageJ without bothering with my python preprocessing! :-) 

> I can only talk for DeepImageJ case: both processings are static IJ macros written for a specific task, then you would suggest adding some key that DeepImageJ understands and the macros as attachments?

Yes, in `config` you can add in as many macro scripts as needed. It would be nice to keep all runtime relevant information in the model.yaml. I would rather not use `attachment` in this scenario as it is consumer specific (e.g. not a pdf documentation that every consumer would be interested in).


> If there are already for the architecture and for the weights, maybe it makes no sense to have it in the root, except if the cards are applications or notebooks for example. It might have more sense to think about how the web-display can be solved first.

I agree and would prefer `model/covers` and `weights/...covers`. The reason to keep it in roots is the way @oeway organized the website and defined valid keys here:
https://github.com/bioimage-io/bioimage.io/blob/master/docs/resource-description-file.md

One of them being `covers`. All these keys are expected in root. I'm fine with either way (if there are only 1+nr_weights `covers` fields in total. Maybe you, @oeway,  @esgomezm and whoever else has an opinion on this can come to a decicion ;-)

oeway:
For the covers, I agree with @FynnBe that we need it for the model architecture, or model in general. On the website, we can also choose to show the covers from the weights when the model covers is missing, but it's better to keep it like that.

About whether we should have it in the root or under model key, I vote for having it in the root to be compatible with the RDF. And try to make the keys under model as minimal as possible, e.g. only contain source and kwargs, and make it optional, so model file with only weights keys + the keys under root (covers, name, description... RDF fields) will serve as a weights file as we discussed. Also not all the model contains source code, some weights file formats (keras h5) contains both the architecture and weights, so the model key can be omitted.

FynnBe:
I don't like the idea of model.yaml files without model source. There is always source code for the architecture, which should be mandatory to specify. A model weights file may only include a traced model, thus (re)training the model is only feasible with the source code.

oeway:
> I don't like the idea of model.yaml files without model source. There is always source code for the architecture, which should be mandatory to specify. A model weights file may only include a traced model, thus (re)training the model is only feasible with the source code.
> 
> edit: Even if the retraining is done manually.

I think we will have the source code for most cases, however, I disagree that source code is absolutely needed. Keras models typically contains the model architecture and the weights, plus if the model is compiled, it also contains the loss, metrics and optimizer states ([ref](https://www.tensorflow.org/guide/keras/save_and_serialize)), which is enough to enable further training. There is also graphical model editor that generate model architecture file (e.g. in JSON or protobuf) that is slightly different from the source code we defined here.

After further digging, I found that ONNX has already implemented feature preview for training support (see [here](https://techcommunity.microsoft.com/t5/azure-ai/onnx-runtime-training-technical-deep-dive/ba-p/1398310)), this means in the future, we will have a single onnx file that can support training.

This moves along the direction of having one file that contains all the information for prediction and retraining.

However, I do agree that ignore the model key and only use `weights` is not a perfect solution. Especially because the key name `weights` doesn't cover model architecture and optimizer state for example.

Maybe we should simply merge `model` and weights key into one?

oeway:
I would like to also add that for most users, we should prefer executing a serialized model without running the source code since executing arbitrary source code downloaded from the internet should not be the default way of getting the model run (mostly for inference). There is a huge security risk for doing that, especially for users dealing with sensitive biomedical data. I understand for now, pytorch doesn't have an easy way to run the model without source code, and there is also risk with only `torch.load` (see [here](https://github.com/pytorch/pytorch/issues/31875)). 

To be clear, I am not saying that we should exclude source code from the model file, one should always attach a github repo with source code to a model, it's just about a potential future direction for us to moving away from the need of running source code. It's might be even better for portability of models, since once the model file format is defined, the library (tensorflow, pytorch) across version can read the same format and perform conversion automatically if needed.

FynnBe:
I think our future plans might get ahead of our current implementation a bit. I would like to suggest to go even a bit simpler and a little more constrained to make it easier for all of us to catch up with the implementation of model runners etc. after this discussion concludes.

To make things easier in the backend:
 - all weights need to have the same format. If not we potentially need to switch runtime environment in a runner (thanks to @m-novikov  for raising this concern) -> new key in root: `weights_format`, remov key in weights: `format`
- as we potentially need the `model/dependencies` for the weighst as well, they have to be in `root` if model is optional -> flatten model into root. new (model) source/sha256 are optional depending on `weights_format` (here having a single `weights_format` already makes it easier to validate the model.yaml)

Additionally I added the `weights/timestamp` field we agreed upon a long time ago (I just had forgotten about it, but found it in the example).

check out the example at https://github.com/bioimage-io/configuration/blob/e2fc8f51f391fc34c0eaa33be8751cca67aa739f/models/UNet2dExample.model.yaml

note: Despite enforcing a one-model.yaml-one-weights-format policy nothing stops us from combining several very similar models into on card on the website.

FynnBe:
On the topic of conversion from old formats, so far our changes are mostly cosmetic and we can map a 0.1/0.2 config to 0.3.0. Once we agree on 0.3.0 I'll add such a mapping to python-bioimage-io

esgomezm:
> * as we potentially need the `model/dependencies` for the weighst as well, they have to be in `root` if model is optional -> flatten model into root. new (model) source/sha256 are optional depending on `weights_format` (here having a single `weights_format` already makes it easier to validate the model.yaml)

It makes sense IMO as the yaml file describes precisely a model. I'm not very sure about the weights format but for this release, it is probably sensible to do it this way. 

Just checking different cases, if we have a model that was trained with a notebook rather than using a GitHub repo, how should we indicate the notebook in the yaml? until now, we used source in root for the notebook and source in `model` for the model architecture.

FynnBe:
> Just checking different cases, if we have a model that was trained with a notebook rather than using a GitHub repo, how should we indicate the notebook in the yaml? until now, we used source in root for the notebook and source in `model` for the model architecture.

How about referencing it in `config/training_notebook`? I see the `config` field as space for those parts of the model that are not (yet) standardized. I am sure we will incorporate keys that some consumers develop under `config` into the common spec over time.

oeway:
For notebooks, in addition to the config field, you can also use the links key when you add your model file to the bioimageio manifest file. This is how the website creat connections across different types of items.

esgomezm:
> For notebooks, in addition to the config field, you can also use the links key when you add your model file to the bioimageio manifest file. This is how the website creat connections across different types of items.

Yes, the link is already being used but then config.yaml wouldn't be self-contained. Config field is fine.

models/UNet2dExample.model.yaml
```diff
@@ -70,53 +50,47 @@ outputs:
...
+    - id: default
+      name: Default Weights
+      description: "Weights from DataScienceBowl"
+      authors: ["author1", "author2"]
+      covers: []
+      source: https://zenodo.org/record/3446812/files/unet2d_weights.torch
+      sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2
+      timestamp: 2019-12-11T12:22:32Z  # ISO 8601
+      test_inputs: my_default_test_input.npy  # language + extension defines memory representation
+      test_outputs: my_default_test_output.npy
```
oeway, line 73: This should be a list

FynnBe: Yes, agreed. We forgot to formalize this when turning the conversation to normalization. I made a separate PR: https://github.com/bioimage-io/configuration/pull/34

PR #36, opened 2020-10-29
[WIP] Tensor Shape Specification

FynnBe:
We have a few open/outdated issues that relate to tensor shape: #13, #17, #14,  #22
Also #12 might be relevant

This PR aims to unite the discussed changes and refine the existing specification and in particular documentation.

PR #42, opened 2020-11-13
Update model spec

constantinpape:
Update the model specification to reflect the changes to model definition and weight definitions.


constantinpape:
I have updated the specification and the unet example now to (what I think) corresponds to the version we agreed upon in the latest meetings.
Let me know if you think there are any additional changes necessary.

I will do a separate PR for the lists of weight_formats and pre/post-processing. 

Also, this superseeds #36.

constantinpape:
@oeway @frauzufall @oeway @FynnBe @m-novikov 
Gentle ping that we should review this to keep track of the changes to the model-spec ;).

README.md
```diff
@@ -21,20 +21,20 @@ Version of this bioimage.io configuration specification. This is mandatory, and
 The recommended behavior for the implementation is to keep backward compatibility, and throw error if the model yaml is in an unsupported format version.
 
 - `name`
-Name of the specification. This name should equal the name of any existing, logically equivalent object of the same category in another language/framework.
+Name of this specification.
```
oeway, line 24: Should this be "Name of this model"?

And perhaps we should define the naming rule. E.g. this should be a human readable name, it should only contain `a-z`, `A-Z`, `-`, `0-9` or space, no more than 36 letters.

README.md
```diff
@@ -63,86 +62,81 @@ Must be a list of *tensor specification keys*.
...
-    - `covers` a list of cover images (see `model`:`covers`). This is used for showing how inputs and outputs look like with this weights file.
-    - `source` link to the weights file. Preferably an url to the weights file.
-    - `sha256` SHA256 checksum of the model weight file specified by `source` (see `models` section above for how to generate SHA256 checksum)
-    - `timestamp` timestamp according to [ISO 8601](#https://en.wikipedia.org/wiki/ISO_8601)
-    - `test_inputs` list of URIs to test inputs as described in `inputs` for a single test case. Supported file formats/extensions: `.npy`
-    - `test_outputs` analog to `test_inputs`.
-    - `documentation` relative path to file with additional documentation in markdown.
-    - `tags` a list of tags.
-    - `attachments` text keys and URI values to additional, relevant files.
+- `attachments` Additional files for this specification; e.g. images that are necessary for the documentation.
```
oeway, line 128: Mention that this will be used for generating the model package.

README.md
```diff
@@ -63,86 +62,81 @@ Must be a list of *tensor specification keys*.
...
+
+- `test_inputs` list of URIs to test inputs as described in inputs for a single test case. Supported file formats/extensions: .npy
+- `test_outputs` analog to test_inputs.
```
oeway, line 131: Is there a `sample_inputs`?

README.md
```diff
@@ -63,86 +62,81 @@ Must be a list of *tensor specification keys*.
...
 
+- `weights` The weights for this model. Weights can be given for different formats, but should otherwise be equivalent.
+   - `weights_format` Format of this set of weights
+        - `authors` a list of authors. This field is optional, only required if the authors are different from the authors specified in root.
+        - `source` link to the weights file. Preferably an url to the weights file.
+        - `sha256` SHA256 checksum of the model weight file specified by `source` (see `models` section above for how to generate SHA256 checksum)
+        - `timestamp` timestamp according to [ISO 8601](#https://en.wikipedia.org/wiki/ISO_8601)
```
oeway, line 138: Would be nice to add `attachments` to allow weights specific attachments for the model packaging, e.g. when we generate a model package with only one weights formats, we will need to know what files to include for this package.

And also perhaps we should mention which field is optional (applies also to the entire doc).

models/UNet2dExample.model.yaml
```diff
@@ -14,17 +14,17 @@ authors:
 cite:
     - text: "Ronneberger, Olaf et al. U-net: Convolutional networks for biomedical image segmentation. MICCAI 2015."
       doi: https://doi.org/10.1007/978-3-319-24574-4_28
+    - text: "2018 Data Science Bowl"
+      doi: https://www.kaggle.com/c/data-science-bowl-2018
```
oeway, line 18:
```suggestion
      url: https://www.kaggle.com/c/data-science-bowl-2018
```

models/UNet2dExample.model.yaml
```diff
@@ -14,17 +14,17 @@ authors:
...
 
 git_repo: https://github.com/bioimage-io/configuration/tree/master/models
-tags: [unet2d, pytorch, nucleus-segmentation]
+tags: [unet2d, pytorch, nucleus-segmentation, data-science-bowl-2018]
```
oeway, line 21: For the display, shorter tags are preferred, how about the following:
```suggestion
tags: [unet2d, pytorch, nucleus, segmentation, dsb2018]
```

models/UNet2dExample.model.yaml
```diff
@@ -59,48 +59,27 @@ outputs:
...
-        - my_other_test_input.npy
-      test_outputs:
-        - my_other_test_output.npy
-      documentation: my_other_weights specific documenation
-      tags: [test, default]
-      attachments: {}  # optional
+test_inputs:  # URI's to inputs (in order as specified in inputs) for a single test case
+  - ./my_default_test_input.npy
+test_outputs:  # analog to test_inputs
+  - ./my_default_test_output.npy
```
oeway, line 71: Add sample_inputs as an example?

oeway (commented):
Made a few comments, thanks for working on this!

README.md
constantinpape: Agreed to all, I will update it.

README.md
constantinpape: Yes.

README.md
constantinpape: Yes, I forgot to add it. Will do so.

README.md
constantinpape: Ok, that use case makes sense. I will add it back here and declare it optional.
I will also try to see if I can spot some other cases where we forgot to mention that it's optional.

models/UNet2dExample.model.yaml
constantinpape: Yeah, good point.
I will see if I can add something.

constantinpape:
@oeway thanks for the comments! I think I have addressed all of them (though I probably overlooked some of the fields that are optional, but we can add this to the spec when we notice it).

models/UNet2dExample.model.yaml
```diff
@@ -5,7 +5,7 @@
 format_version: 0.3.0
 
 name: UNet2dExample
```
oeway, line 7: Perhaps better to add a space to make it more human readable?

oeway (approved):
Looks good!

With a minor comment about the example model name.

models/UNet2dExample.model.yaml
constantinpape: I used a dash now ;)

README.md
```diff
@@ -63,86 +62,85 @@ Must be a list of *tensor specification keys*.
...
+
+- `test_inputs` list of URIs to test inputs as described in inputs for a single test case. Supported file formats/extensions: .npy
+- `test_outputs` analog to test_inputs.
+
+- `sample_inputs` list of URIs to sample inputs to illustrate possible inputs for the model, for example stored as png or tif images. This field is optional. 
+- `sample_outputs` list of URIs to sample outputs corresponding to the `sample_inputs`. This field is optional.
 
+- `weights` The weights for this model. Weights can be given for different formats, but should otherwise be equivalent.
+   - `weights_format` Format of this set of weights
+        - `authors` a list of authors. This field is optional, only required if the authors are different from the authors specified in root.
```
frauzufall, line 138: the indentation looks off here

README.md
```diff
@@ -63,86 +62,85 @@ Must be a list of *tensor specification keys*.
...
-    - `attachments` text keys and URI values to additional, relevant files.
+- `attachments` Additional files for this specification; e.g. images that are necessary for the documentation. These files will be included when generating the model package. This field is optional.
+
+- `test_inputs` list of URIs to test inputs as described in inputs for a single test case. Supported file formats/extensions: .npy
+- `test_outputs` analog to test_inputs.
+
+- `sample_inputs` list of URIs to sample inputs to illustrate possible inputs for the model, for example stored as png or tif images. This field is optional. 
+- `sample_outputs` list of URIs to sample outputs corresponding to the `sample_inputs`. This field is optional.
 
+- `weights` The weights for this model. Weights can be given for different formats, but should otherwise be equivalent.
```
frauzufall, line 136: should we maybe add `kwargs` to a weights entry? E.g. for the TensorFlow SavedModelBundle one has to specify a `tag` when loading the graph (which is commonly `serve`). What's in the `kwargs` could be specified depending on the `weights_format`.

frauzufall (commented):
I added two more comments and I have another question: One might want to save the TensorFlow version which was used to produce a `weights` entry - where would one do that?

README.md
oeway: I think you can define your own required/optional keys contained in each weights format, we just need to document them in each weight format definition.

So in that case, I would directly add a `tag` key under the weight format, then document that the `tag` key is required for your tensorflow format.

README.md
constantinpape: I think it's corrected, it's nested 3 times; this is how it renders on github:

indent_model

README.md
constantinpape: I think the suggestion by @oeway is good to have custom keys for the weight formats and then have `tag` as a requirement for the tf weight format.

constantinpape:
> I added two more comments and I have another question: One might want to save the TensorFlow version which was used to produce a `weights` entry - where would one do that?

Thanks for your comments @frauzufall.

About the TensorflowVersion: could we save it as another (maybe optional) field in the tensorflow weight format, following the suggestion by @oeway?

README.md
constantinpape: If we agree on this, I will add some description saying that additional fields for the different formats are allowed and we can work on defining the formats in #43.

README.md
frauzufall: Ah, it's a map, not a list! Got it.

frauzufall (approved):

constantinpape:
Thanks for the reviews @frauzufall and @oeway. I will go ahead and merge this.

PR #49, opened 2020-11-25
Tables update

FynnBe:


README.md
```diff
@@ -65,17 +65,12 @@ Must be a list of *tensor specification keys*.
...
     or (only for output) as {`reference_input` *input tensor name*, `scale` *list of factors 'output_pix/input_pix' for each dimension*, `offset` *position of origin wrt to input*}
-  - `[preprocessing]` (only for input) optional list of transformations describing how this input should be preprocessed. Each entry consists of these keys:
-    - `name` name of preprocessing (currently only 'zero_mean_unit_variance' is supported)
-    - `[kwargs]` optional key word arguments for `preprocessing`\
-        for 'zero_mean_unit_variance' these are:
-        - `mode`: either 'fixed', 'per_dataset', or 'per_sample'
-        - `axes`: subset of axes to normalize jointly, e.g. 'xy', batch ('b') is not a valid axis key here!
-        - `mean`: mean if mode == fixed, e.g. (with channel dimension of length c=3, and all axes 'cxy') [1.1, 2.2, 3.3]
-        - `std`: standard deviation if mode == fixed analogously to mean
+  - `[preprocessing]` (only for input) optional list of transformations describing how this input should be preprocessed. Each entry has these keys:
```
constantinpape, line 68: Should we remove the 'optional' here as well, because it's clear due to the brackets? And same for postprocessing below.

constantinpape (commented):
Looks good. Do you want to keep mentioning in the description that fields are optional if they are marked by `[]`? I think it might be good to remove it from the text then to be less dense.

README.md
```diff
@@ -92,9 +87,9 @@ The deep learning framework of the source code. For now, we support `pytorch` an
 Can be `null` if the implementation is not framework specific.\
 `language` and `framework` define which model runner can use this model for inference. 
 
-- `source`
+- `[source]`
```
FynnBe, line 90: here, it's only optional depending on `weights_format`. should we indicate this differently? `[...]*` *depends on `other_field` ?

constantinpape: Yeah, I think `[...]*` is good. (And in that case we should always explain it more in the description like here.)

constantinpape:
@FynnBe why did you remove the local example? I think it's good to have it here for reference.

FynnBe:
> @FynnBe why did you remove the local example? I think it's good to have it here for reference.

maintenance mainly. The spec, the reade, the example and the 'real example' in pybio.torch are just a lot of places to keep in sync. It's a single click away, either way. I think it is better to have a working example (that passes a real test) than just an unchecked additional yaml with no guarantees.

constantinpape:
Ah, sorry I was unaware of the version in pybio.torch and I missed the change in the readme where you link to it.
I agree, it's good to remove it here then.

constantinpape (approved):
I think this is good to go now. Let me know if you want to change anything else, @FynnBe.

FynnBe:
LGTM, always room for another PR, but this one should come to an end 👍

Issue #48, opened 2020-11-24
Model framework


frauzufall:
Hi group, what does `framework` mean in the current version of the model specification? Is it the framework which was used for training? Or is it obsolete because we can now support multiple frameworks by having multiple weights of different types? Or is it a a class bundling different types of convertible weights, e.g. `tensorflow` for `tensorflow_saved_model_bundle` and `tensorflow_js`?

constantinpape:
It is the framework used for training, so e.g. `pytorch` or `tensorflow`. When we initially came up with the spec, this was also the only framework that would be supported for prediction too.

Obviously this changed now that we support weights in multiple formats, but I think we should still leave it in, but maybe add some more description to make this clear?

FynnBe:
It refers to the framework of the source code (so does language), and thus not only limited to a future training, but also relevant for any weights for inference that depend on source code.

constantinpape:
I do think we still need these fields, but the description and restrictions we currently have in the doc don't make much sense any more. Here's the current version:

```
- `language`
Programming language of the source code. For now, we support `python` and `java`.
<!---
What about `javascript`?
-->

- `framework`
The deep learning framework of the source code. For now, we support `pytorch` and `tensorflow`.
Can be `null` if the implementation is not framework specific.\
`language` and `framework` define which model runner can use this model for inference. 
```

The biggest issue is that 
> `language` and `framework` define which model runner can use this model for inference. 

is not true any more. The available runner(s) now depend on the weight format(s). 

Additionally, I am not really sure we even need to restrict the programming language. As long as a valid model drops out of it we don't care (except for some weight formats that expect the source code in a certain language in order to get the model definition from it).

FynnBe:
let's remove the part about the runner then, but imho we should keep a strict list of languages (and frameworks). Otherwise someone might just have a typo in 'pyton' and then 'pytorch_state_dict' weight's can't be loaded. If anyone has model source code in a different langauge, it'd be easy to add it.

constantinpape:
> Otherwise someone might just have a typo in 'pyton' and then 'pytorch_state_dict' weight's can't be loaded. If anyone has model source code in a different langauge, it'd be easy to add it.

Ok, that's a good point, we can easily extend the allowed languages if necessary.

constantinpape:
I thought about this further, and I think that `language`, `framework` and `kwargs` have to be `[...]*` as well, because they are not required (and don't even make sense) if `source` is not present.
[...]

PR #50, opened 2020-11-26
Mark fields depending on source as optional *

constantinpape:


constantinpape:
As discussed in #48, `language` and `framework` should be `*optional`, because they only need to be included if we have `source` (which in turn is only required for some of the weight formats).

FynnBe (approved):

FynnBe:
closes https://github.com/bioimage-io/configuration/issues/53

constantinpape:
Pending #54

PR #51, opened 2020-11-27
Add execution_model field

constantinpape:
As discussed in the meeting, this adds the `execution_model` field to specify custom prediction procedutes.
Let me know if you agree with the description here and if we want to stick wit the name or still change it.
(I am just putting a veto on `runner`, because it's ambiguous with our other usage of this term ;))

@oeway @frauzufall @fjug @esgomezm @FynnBe @m-novikov

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,9 +138,9 @@ For example:
         - `sha256` SHA256 checksum of the model weight file specified by `source` (see `models` section above for how to generate SHA256 checksum)
         - `[attachments]` weight specific attachments that will be included when generating the model package.
 
-- `[execution_model]` Custom mode for running prediction with this model. For more complex prediction procedures like test time data augmentation that cannot be expressed in the current model configuration. The different execution models should be listed in [supported_formats_and_operations.md#Execution Model](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#execution_model)
-  - `name` the name of the execution model
-  - `[kwargs]` keyword arguments for this execution model
+- `[run_mode]` Custom run mode for this model: for more complex prediction procedures like test time data augmentation that currently cannot be expressed in the specification. The different run modes should be listed in [supported_formats_and_operations.md#Run Modes](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#run modes)
+  - `name` the name of the `run_mode`
+  - `[kwargs]` keyword arguments for this `run_mode`
  
 - `[config]`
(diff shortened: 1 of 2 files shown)


esgomezm:
Thank you!
Now, reading the description (which is good for me), I do agree with @FynnBe that seems more descriptive the "how to use" name. Otherwise, I would maybe write "model_execution".

README.md
```diff
@@ -137,6 +137,10 @@ For example:
         - `source` link to the weights file. Preferably an url to the weights file.
         - `sha256` SHA256 checksum of the model weight file specified by `source` (see `models` section above for how to generate SHA256 checksum)
         - `[attachments]` weight specific attachments that will be included when generating the model package.
+
+- `[execution_model]` Custom mode for running prediction with this model. For more complex prediction procedures like test time data augmentation that cannot be expressed in the current model configuration. The different execution models should be listed in [supported_formats_and_operations.md#Execution Model](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#execution_model)
```
oeway, line 141: The explanation seems suggest another key name `execution_mode`, otherwise we need to explain it is a model for execution of the current model ;)

constantinpape: Yes, I agree. `execution_mode` makes more sense.

FynnBe:
I am fine with `execution_mode`, *but* 😇 still want to throw `how_to_run` as a self-explanatory field name in the race, pointing at the runner that it is directly related to, without actually including `runner`, which clearly distinguishes it. Alternatively `run_mode` makes more sense to me (or we rename the infamous runner to executor, which has that slightly morbid/generic ring to it).

constantinpape:
Ok, I made a poll with all the options listed here:

[](https://api.gh-polls.com/poll/01ERC4AXAANQ619JSQEVFP8VVF/execution_mode/vote)
[](https://api.gh-polls.com/poll/01ERC4AXAANQ619JSQEVFP8VVF/how_to_run/vote)
[](https://api.gh-polls.com/poll/01ERC4AXAANQ619JSQEVFP8VVF/run_mode/vote)
[](https://api.gh-polls.com/poll/01ERC4AXAANQ619JSQEVFP8VVF/model_execution/vote)

FynnBe:
> (I think you can vote multiple times, but please vote only once ;))

only once per option, though?!?

constantinpape:
> > (I think you can vote multiple times, but please vote only once ;))
> 
> only once per option, though?!?

To be more precise: although the simple polling implementation from https://github.com/apex/gh-polls allows you to vote as much as you want, please vote only one time (in total) ;).

constantinpape:
aftter discussion in the bioimage.io meeting `run_mode

Issue #41, opened 2020-11-10
Is it necessary to have the model name in <model name>.model.yaml?


frauzufall:
Hi all,

Is the intention behind naming the specification file `<model name>.model.yaml` that there might be multiple of them in the same ZIP?

In our case the ZIP content is cached somewhere (imagine it's located remotely) which makes it a bit difficult to scan the whole content for a `*.model.yaml` file, I'd rather like to know exactly for which file name to ask. 

This is a problem with the API on our side, but I want to make sure I don't misunderstand the use case as well. Therefore my question - why is the file not simply named `model.yaml` or if there is the possibility for multiple YAML files, should there be something like an index file listing the available model files in this ZIP?

Thanks for pointers!
Deborah

oeway:
Hi,

How about allowing only one model in the zip package and rename the file name as `model.yaml`? 

For the naming of the yaml file in a github repo, we have basically two choices:

1. force each model yaml to be contained in a folder so we have the model:
```
+-- my-awesome-model-1
|   +-- model.yaml
|   +-- my-other-files.xyz
+-- my-awesome-model-2
|   +-- model.yaml
|   +-- my-other-files.xyz
```
The drawback of this is that if you want to share files (e.g. source code) between models, you will need to copy them (resolving relative path `../` won't be a good option for packaging).

2. use any name ends with `.model.yaml`, and only rename them into `model.yaml` when they are packaged into a zip file. And this process can be automised from the bioimage.io website with packaging bioengine app. Then we can have this type of structure:
```
+-- my-awesome-models
|   +-- my-awesome-model-1.model.yaml
|   +-- my-awesome-model-2.model.yaml
|   +-- my-other-files.xyz
```
After packaging, one of the zip package will become:
```
+-- my-awesome-model-1.zip
|   +-- model.yaml
|   +-- my-other-files.xyz
```

oeway:
> Should the .zip file have the .model.zip extension (to avoid future clashes with other packages we might want to publish, e.g. .dataset.zip/ .notebook.zip or whatever? Otherwise we again are in the position of having to look into the zip to determine what kind of package it is.

I think using `.model.zip` is a good idea!

frauzufall:
[...]

Issue #22, opened 2020-03-24
Real size of the output (definition of scale and offset)


esgomezm:
While trying to adapt DeepImageJ to the YAML file, we had some doubts about how to specify the shape of the output. Here I let you the questions we come up with: 
- Is there any specific definition for scale and offset when defining the shape of the output?  
- What is the reference for `scale`? Is it `output_shape = input_shape * scale` or `input_shape = output_shape*scale`?
- For the `offset`, does it refer to difference between the shape of the input and the output? after or before the scaling? What is the main difference between `halo` and `offset`?

If we want to do tiling, we should consider both the size of input and output tiles, and the size of the final output after tiling. A good example could be doing super-resolution (the output is larger than the input) with an encoder-decoder. Is there any consensus about this? If not, I suggest the following:

- Shape of the model's output:
`output_shape = input_shape - offset`

- Shape of the receptive field of the output (of the output tile): 
`receptive_field_shape = output_shape - 2*halo`

- Shape of the resulting image: 
`result_shape = (tamaño_imagen)*scale`

oeway:
@FynnBe and @constantinpape may give clarification for these question, but I started to think is there "smarter" way of doing the shape calculation. 

For example, instead of defining scale, offset, halo, how about allow a math equation, say, `output_shape: 8*(input_shape*2) + 31`? Or even point to a function just like the preprocessing transforms?

constantinpape:
>     * What is the reference for `scale`? Is it `output_shape = input_shape * scale` or `input_shape = output_shape*scale`?

It is `output_shape = input_shape * scale + offset`.
Tou can think about `output_shape` (`y`) being a linear function of the `input_shape` (`x`):
```
y = a * x + b
```
Then `a = scale` and `b = offset`.

>     * For the `offset`, does it refer to difference between the shape of the input and the output? after or before the scaling? What is the main difference between `halo` and `offset`?

For `offset` see above. `halo` is a parameter for prediction and specifies how much if the predicted tensor should be cropped in order to avoid border artifacts.
E.g. if you have a network that returns images of size `[256, 256]` and you specify `halo = [32, 32]`,
this tells the application that runs prediction to crop away 32 pixels on each side; so the output will be
`[256 - 2 * 32, 256 - 2 * 32] = [192, 192]`.

I think this covers all the use cases you describe lower.

@oeway

> For example, instead of defining scale, offset, halo, how about allow a math equation, say,

[...]

Issue #14, opened 2019-12-16
Is null necessary in input / output shape step?


ctrueden:
In the `shape` of `inputs` and `outputs`, the `step` can have `null` as a value. Does this mean 0? Or 1? Or something else? If it means 0 or 1, consider getting rid of it?

oeway:
I think we can make the step field optional and it must be a number. I agree that If the value is null, we should get rid of it.

constantinpape:
I think right now `null` means that we don't allow for any `step` (is this correct @FynnBe  ?).
I agree that it's redundant in this case abd we should not allow for `null`.

esgomezm:
> I think right now `null` means that we don't allow for any `step` (is this correct @FynnBe  ?).
> 
Could it also be possible to consider `null` to allow any `step`? Maybe you are already referring to similar cases: if the network is an encoder, and has a minimum size but not fixed, then, the entire image could be the input for the network, so the step would be the largest as possible. 

Is there any other case in which you would not allow any `step` and the `shape` is not `exact`?

FynnBe:
if there is a fixed shape it should set as `shape`:
```
shape: [1, 2, 3, 4]
```
if there is a minimum and a step:
```
shape:
  minimum: [1, 2, 3, 4]
  step: [1, 4, 6, 8]
```  
or to allow any step:
```
shape:
  minimum: [1, 2, 3, 4]
  step: [1, 1, 1, 1]
```  
if it's more complicated than that:
```
shape: dynamic
```
and the object should have a method `dynamic_output_shape`

if the whole list of tensors cannot be described:
```
inputs/outputs: dynamic
```
and the object should have a method `dynamic_inputs`/`dynamic_outputs`

(as of `format_version` 0.1.0)
sorry this is not fully documented yet

deepimagej:
> if there is a fixed shape it should set as `shape`:
> 
> ```
> shape: [1, 2, 3, 4]
> ```
So we do not use `exact` as described in the documentation?

Ok for the rest. Thank you very much!

FynnBe:
yes, that drop out in the current version 0.1.0, sorry it's still ghosting around

Carloscj18:
What if only one dimension of the shape is fixed, let's say the last one. Could it be like the following?
shape:
  minimum: [1, 2, 3, 4]
  step: [1, 4, 6, 0]

Where 0 implies there is no possibility to change that dimension.

[...]

PR #54, opened 2020-11-27
Extend description

esgomezm:
Main changes:
- Name of the [model.yaml](https://github.com/bioimage-io/configuration/compare/master...esgomezm:master#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R15) according to #41.
- output shape description as discussed at #22 and #14.


README.md
```diff
@@ -88,7 +98,7 @@ What about `javascript`?
 - `framework`
 The deep learning framework of the source code. For now, we support `pytorch` and `tensorflow`.
 Can be `null` if the implementation is not framework specific.\
-`language` and `framework` define which model runner can use this model for inference. 
+`language` and `framework` define which model runner can use this model for inference (see [supported_formats_and_operations.md#preprocessing](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#preprocessing)).
```
constantinpape, line 101: This part of the spec is actually not quite correct, and I proposed changes to it in #50.

README.md
```diff
@@ -133,13 +143,13 @@ For example:
 
 - `weights` The weights for this model. Weights can be given for different formats, but should otherwise be equivalent.
    - `weight_format` Format of this set of weights. Weight formats can define additional (optional or required) fields. See [supported_formats_and_operations.md#Weight Format](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#weight_format)
-        - `[authors]` a list of authors.
+        - `[authors]` a list of authors. It refers to the ones who trained the model.
```
constantinpape, line 146: I think `It refers to the ones who trained the model` is not quite correct here, because the optional `authors` field here indicates to who create this set of weights, which could also happened by converting the weights from one format to another (so not by training the model).

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
...
+    - `min` *minimum shape with same length as `axes`*
+    - `step` *minimum shape change with same length as `axes`*\
+    
+    or (only for output) as\
+    `shape`
+    - `reference_input` *input tensor name*
+    - `scale` *list of factors 'output_pix/input_pix' for each dimension*
+    - `offset` *position of origin wrt to input*
+    - `halo` *the halo to crop from the output tensor* (for example to crop away boundary effects or for tiling) (e.g. the valid domain of an output of `size = [256, 256]` and `halo = [32, 32]`, results in `output_shape = [256 - 2 * 32, 256 - 2 * 32] = [192, 192]`).\
+    The output shape is determined as `output_shape = reference_input_shape * scale + offset`.
```
constantinpape, line 80: This is great! The old layout was very confusing.

README.md
constantinpape: I am not sure how to best combine these two PRs, maybe it's best if we merge this one and then just rebase #50 on top of it.

supported_formats_and_operations.md
```diff
@@ -40,8 +40,8 @@ The supported preprocessing operations.
   - `kwargs`
     - `mode` can be one of `per_sample` (percentiles are computed for each sample individually), `per_dataset` (percentiles are computed for the entire dataset)
     - `axes` the subset of axes to normalize jointly. For example `xy` to normalize the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
-    - `min_percentile` the lower percentile used for normalization, in range 0 to 100.
-    - `max_percentile` the upper percentile used for normalization, in range 0 to 100. Has to be bigger than `upper_percentile`.
+    - `min_percentile` the lower percentile used for normalization, in range 0 to 1
+    - `max_percentile` the upper percentile used for normalization, in range 0 to 1. Has to be bigger than `upper_percentile`.
```
constantinpape, line 44: percentile usually refers to a range from 0 to 100, not 0 to 1.
See https://en.wikipedia.org/wiki/Percentile.
I think it would be confusing if we change this from the common definition.
@frauzufall how do you handle this in `CSBDeep`?

constantinpape (commented):
Thanks for working on this @esgomezm.
Changes look mostly good, we should just clarify the definition of percentile.

README.md
esgomezm: Sorry, I missed that one. Then, in the `weights` or `weight_format` definition, I'd indicate that this field is what determines `which model runner can load and use this model for inference`. I'd say that this is a key detail for future contributors. For example:
```yaml
- weight_format` Format of this set of weights. Each model runner is limited to a specific format. Weight formats can define additional (optional or required) fields. See [supported_formats_and_operations.md#Weight Format](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#weight_format)
```

> I am not sure how to best combine these two PRs, maybe it's best if we merge this one and then just rebase #50 on top of it.

Go on with the easiest way for you. I can also remove this PR and do it on a different branch.

README.md
esgomezm: Sorry, I thought those authors were the ones in the main description as it says
`The authors are the creators of the specifications and the primary points of contact.`. I would expect that the ones who created the specification are also the ones responsible for the weights format. 
Maybe we could agree on a general and more specific description?

supported_formats_and_operations.md
esgomezm: Yes, this was a suggestion coming from [#43](https://github.com/bioimage-io/configuration/pull/43#issuecomment-733554840). On one side I agree that percentiles are quantiles, but on the other, it does really depend on the software you use, and as @FynnBe suggested, some cases can be ambiguous.

README.md
constantinpape: > Sorry, I missed that one. Then, in the `weights` or `weight_format` definition, I'd indicate that this field is what determines `which model runner can load and use this model for inference`

Yes, I fully agree. We should add this to `weights` or `weight_format`.

> Go on with the easiest way for you. I can also remove this PR and do it on a different branch.

Let's go ahead with this one and I will rebase #50 later.

README.md
constantinpape: So the idea is that `authors` in the main one are the ones that created the model and uploaded the `model.yaml` etc.
After that happens someone else might convert the weights to a different format and then add this to the same `model.yaml`. In this case, they should be listed here (because they are different from the "main" authors).

supported_formats_and_operations.md
constantinpape: No, @FynnBe did not suggest to change the value range to `[0, 1]`. Instead he suggested to enforce `max_percentile > 1` in order to avoid confusion about the value ranges.

README.md
constantinpape: (I agee, we should explain this in more detail)

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
   *tensor specification keys*:
   - `name` tensor name
   - `data_type` data type (e.g. float32)
-  - `data_range` tuple of (minimum, maximum)
-  - `axes` string of axes identifying characters from: btczyx
+  - `data_range` tuple of (minimum, maximum) (e.g. [-inf, inf])
```
FynnBe, line 65: Just a hunch: Did you remove the inf example, because you expect `data_range` to hold the actual minimum, maximum of the tensor?
I think for now it was intended as the threoretical data_range (that might be given by the `data_type`), which the example clarifies. (a comment might be clearer, though). In combination with the data type this is of limited use, but the actual (minimum,maximum) would confusing due to the batch dimension, which is arbitrary and thus has a random influence on these values. The actual, global data range might not be required, thus unnecessary to be computed, or not sufficient (e.g. for percentile normalization). I suggest not to change anything further for 0.3.0, and keep it as theoretical data range.

README.md
```diff
@@ -12,7 +12,7 @@ To get a quick overview of the config file, see an example file [here](https://g
 
 ## Model Specification
 
-A model entry in the bioimage.io model zoo is defined by a configuration file `<model name>.model.yaml`.
+A model entry in the bioimage.io model zoo is defined by a configuration file `model.yaml`.
```
FynnBe, line 15:
```suggestion
A model entry in the bioimage.io model zoo is defined by a configuration file `<optional model name>.model.yaml`. Once a model is packaged to a `<(mandatory) model name>.model.bioimage.io.zip` file the configuration file inside the .zip is only named `model.yaml`.
```
The model entry is not the same as the packaged model (where we agreed to `model.yaml` inside a `<model name>.model.bioimage.io.zip`, see https://github.com/bioimage-io/configuration/issues/41). Keeping several models in one folder should be possible, as the may share other resources, etc.

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
...
+  - `axes` string of axes identifying characters from: btczyx. Same length and order as the axes in `shape`.
   - `shape` specification of tensor shape\
-    Either as *exact shape with same length as `axes`*,\
-    or (only for input) as {`min` *minimum shape with same length as `axes`*, `step` *minimum shape change with same length as `axes`*},\
-    or (only for output) as {`reference_input` *input tensor name*, `scale` *list of factors 'output_pix/input_pix' for each dimension*, `offset` *position of origin wrt to input*, `halo` *the halo to crop from the output tensor (for example to crop away boundary efects or for tiling)*}
+    Either as *fixed shape with same length as `axes`* (e.g. `shape`: [1, 512, 512, 1]),\
+    or (only for input) as\
+    `shape`
+    - `min` *minimum shape with same length as `axes`*
+    - `step` *minimum shape change with same length as `axes`*\
```
FynnBe, line 72:
```suggestion
    - `step` *minimum shape change with same length as `axes`*
```
remove trailing \

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
...
-  - `data_range` tuple of (minimum, maximum)
-  - `axes` string of axes identifying characters from: btczyx
+  - `data_range` tuple of (minimum, maximum) (e.g. [-inf, inf])
+  - `axes` string of axes identifying characters from: btczyx. Same length and order as the axes in `shape`.
   - `shape` specification of tensor shape\
-    Either as *exact shape with same length as `axes`*,\
-    or (only for input) as {`min` *minimum shape with same length as `axes`*, `step` *minimum shape change with same length as `axes`*},\
-    or (only for output) as {`reference_input` *input tensor name*, `scale` *list of factors 'output_pix/input_pix' for each dimension*, `offset` *position of origin wrt to input*, `halo` *the halo to crop from the output tensor (for example to crop away boundary efects or for tiling)*}
+    Either as *fixed shape with same length as `axes`* (e.g. `shape`: [1, 512, 512, 1]),\
+    or (only for input) as\
```
FynnBe, line 69:
```suggestion
    or (only for input) as
```
remove trailing \

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
...
+    `shape`
+    - `min` *minimum shape with same length as `axes`*
+    - `step` *minimum shape change with same length as `axes`*\
+    
+    or (only for output) as\
+    `shape`
+    - `reference_input` *input tensor name*
+    - `scale` *list of factors 'output_pix/input_pix' for each dimension*
+    - `offset` *position of origin wrt to input*
+    - `halo` *the halo to crop from the output tensor* (for example to crop away boundary effects or for tiling) (e.g. the valid domain of an output of `size = [256, 256]` and `halo = [32, 32]`, results in `output_shape = [256 - 2 * 32, 256 - 2 * 32] = [192, 192]`).\
```
FynnBe, line 79: In my perception the `halo` border is actually not cropped away, but returned. However, your description of `halo` makes more sense. We should never return anything invalid. I saw a problem with 'same' convolutional models, that do return the same output size but have a certain halo of (more or less) invalid border data. I suppose those models should incorporate a postprocessing crop and only return valid data, turning the same convolution model into a valid one. @constantinpape ,@m-novikov, what do you think?

README.md
```diff
@@ -62,12 +62,22 @@ Must be a list of *tensor specification keys*.
...
+    - `min` *minimum shape with same length as `axes`*
+    - `step` *minimum shape change with same length as `axes`*
+    
+    or (only for output) as\
+    `shape`
+    - `reference_input` *input tensor name*
+    - `scale` *list of factors 'output_pix/input_pix' for each dimension*
+    - `offset` *position of origin wrt to input*
+    - `halo` *the halo to crop from the output tensor* (for example to crop away boundary effects or for tiling) (e.g. the valid domain of an output of `size = [256, 256]` and `halo = [32, 32]`, results in `output_shape = [256 - 2 * 32, 256 - 2 * 32] = [192, 192]`).\
+    The output shape is determined as `output_shape = reference_input_shape * scale + offset`.
```
FynnBe, line 80:
```suggestion
    The output shape is determined as `output_shape = reference_input_shape * scale - 2*halo.
```
in the description of `offset` it says it describes the relative position to the reference input, slighly chaning the definition of halo (really just clarifying it's meaning, see above, halo should be subtracted twice, instead

README.md
constantinpape: I think putting in the "data range so that the model yields sensible results" makes sense for the data range:
https://github.com/bioimage-io/configuration/issues/55#issuecomment-735646024

README.md
constantinpape: > In my perception the `halo` border is actually not cropped away, but returned.

But what's the difference then of having a halo or not?

The way I understand halo here is that it's a hint to the consumer runing prediction with this model, that indicates which "outer region" of the prediction should be discarded, for example due to boundary artifacts caused by same convolutions. (but there could be lots of other reasons for doing this)

FynnBe: > But what's the difference then of having a halo or not?

I am confused, as to me you answer this question yourself:

>The way I understand halo here is that it's a hint to the consumer running prediction with this model, that indicates which "outer region" of the prediction should be discarded, for example due to boundary artifacts caused by same convolutions. (but there could be lots of other reasons for doing this)

This is how I understood it, too. I have to say though, that I like @esgomezm definition more that forces you to crop the outputs that contain a halo (either through valid convs or postprocessing), to instead return only valid output. Every consumer software should just tile accordingly and would no longer have to deal with the two options of 1. I ignore the halo and look at ugly border artefacts or 2. I crop the halo from what I got back from the runner. It would amount to option 3: I have to tile correctly (overlapping if model has halo), but what I get back is all valid :-)

constantinpape: Ah ok, so now I understand the issue here, I think.
I think we all agree on the principle definition of halo, but the question is who's responsible for cropping it. 
1. It's cropped in the mode already.
2. It's cropped by the consumer.

I don't think option 1 makes so much sense, because then we wouldn't need to specify the halo in the first place, because we can just use the predictions from the model in the first place.

But if we go with option 2, this implies that we don't change the output shape when a halo is specified.

README.md
constantinpape: Ok, I actually think we shouldn't include the halo in the `output_shape`, see https://github.com/bioimage-io/configuration/pull/54#discussion_r532500407.
Maybe we should make a separate issue for this discussion...

README.md
FynnBe: > I don't think option 1 makes so much sense, because then we wouldn't need to specify the halo in the first place, because we can just use the predictions from the model in the first place.

right, if we do want to handle it this way, we could just move it into output `shape`, as in this case it's only applicable when calculating the actual output from a reference input

README.md
constantinpape: On second thought, I am not sure we should mention this here, it's difficult to understand this without explaining the `...model.zip` (which we are not doing yet, but need to add here as well.)

constantinpape:
Closing this PR in favor of #59

Issue #58, opened 2020-12-04
Add description of "model.bioimage.io.zip" to the README


constantinpape:
We are currently not describing the zip we use to bundle the model.

FynnBe:
We have documented this in a few places by now, e.g. [here](https://bioimage-io.github.io/spec-bioimage-io/v0.5.14.1/api/#bioimageio.spec.save_bioimageio_package)

PR #59, opened 2020-12-04
Shape description and authors

constantinpape:
This PR includes:
- extension of the shape description based on the work by @esgomezm in #54 
- update to the `author` fields and introduction of the `parent` fields

Please have a look if this reflects what we agreed upon in the meeting today and feel free to suggest changes, @esgomezm @oeway @frauzufall.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The person(s) that have packaged and uploaded this model. Only needs to be speci
 
 - [`parent`] Parent model from which this model has been derived, e.g. by finetuning the weights of this model on a different dataset.
   - `uri` Url of another model available on bioimage.io or path to a local model in the bioimage.io specification. If it is a url, it needs to be a github url linking to the page containing the model (NOT the raw file). 
-  - `hash` hash of the weights of the parent models
+  - `sha256` hash of the weights of the parent model.
 
 - `inputs`
 Describes the input tensors expected by this model.
@@ -68,8 +68,8 @@ Must be a list of *tensor specification keys*.
 
   *tensor specification keys*:
   - `name` tensor name
-  - `data_type` data type (e.g. float32)
-  - `data_range` tuple of (minimum, maximum), e.g. [0, 1]
+  - `data_type` the data type of this tensor. For inputs, only `float32` is allowed and the consumer software needs to ensure that the correct data type is passed here. For outputs can be any of `float32, float64, (u)int8, (u)int16, (u)int32, (u)int64`. The data flow in bioimage.io models is explained [in this diagram.](https://docs.google.com/drawings/d/1FTw8-Rn6a6nXdkZ_SkMumtcjvur9mtIhRqLwnKqZNHM/edit).
+  - `[data_range]` tuple `(minimum, maximum)` specifying the allowed range of the data in this tensor. If not specified, the full data range that can be expressed in `data_type` is allowed.
   - `axes` string of axes identifying characters from: btczyx. Same length and order as the axes in `shape`.
   - `shape` specification of tensor shape. It can be specified in three diffeent ways:
     1. as exact shape with same length as `axes`, e.g. `shape: [1, 512, 512, 1]`
--- a/README.md
+++ b/README.md
@@ -55,6 +55,13 @@ Relative path to file with additional documentation in markdown.
 - `attachments`
 Dictionary of text keys and URI values to additional, relevant files.
 
+- [`pacakged_by`]
+The person(s) that have packaged and uploaded this model. Only needs to be specified if different from `authors` in the root weights`, see `weights` for more details.
+
+- [`parent`] Parent model from which this model has been derived, e.g. by finetuning the weights of this model on a different dataset.
+  - `id` Url of another model available on bioimage.io or path to a local model in the bioimage.io specification. In addition to the 
+  - `hash` hash of the weights of the parent models
+
 - `inputs`
 Describes the input tensors expected by this model.
 Must be a list of *tensor specification keys*.
@@ -144,7 +151,8 @@ For example:
 
 - `weights` The weights for this model. Weights can be given for different formats, but should otherwise be equivalent. The available weight formats determine which consumers can use this model.
    - `weight_format` Format of this set of weights. Weight formats can define additional (optional or required) fields. See [supported_formats_and_operations.md#Weight Format](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#weight_format)
-        - `[authors]` a list of authors.
+        - `authors` a list of authors. If this is the root weight (it does not have a `parent` field): the person(s) that have trained this model. If this is a child wieght (it has a `parent` field): the person(s) who have converted the weights to this format.
+        - [`parent`]* the source weights used as input for converting the weights to this format. For example, if the weights were converted from the format `pytorch_state_dict` to `pytorch_script`, the parent is `pytorch_state_dict`. All weight entries except one (the initial set of weights resulting from training the model), need to have this field.
         - `source` link to the weights file. Preferably an url to the weights file.
         - `sha256` SHA256 checksum of the model weight file specified by `source` (see `sha256` section for how to generate the checksum)
         - `[attachments]` weight specific attachments that will be included when generating the model package.
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Must be a list of *tensor specification keys*.
        - `reference_input` name of the reference input tensor
        - `scale` list of factors 'output_pix/input_pix' for each dimension
        - `offset` position of origin wrt to input 
-       - `halo` the halo to crop from the output tensor (for example to crop away boundary efects or for tiling). The halo should be croped from both size, i.e. `shape_after_crop = shape - 2 * halo`.
+       - `halo` the halo to crop from the output tensor (for example to crop away boundary efects or for tiling). The halo should be croped from both size, i.e. `shape_after_crop = shape - 2 * halo`. The `halo` is not cropped by the bioimage.io model, but is left to be cropped by the consumer software. See `offset` for a 'halo' that is cropped by the model.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ To get a quick overview of the config file, see an example file [here](https://g
 
 ## Model Specification
 
-A model entry in the bioimage.io model zoo is defined by a configuration file `<model name>.model.yaml`.
```
FynnBe, line 15: isn't the `<model name>` optional, as in it might be there?

oeway: Not exactly, we said that if the yaml file is inside a model package, then we should remove the model name part, but if it's outside, e,g. on github, we should have it.

constantinpape: Yes, I thought it was better to remove `<model name>` for now, because we haven't defined the differences between the config on github vs in the zip bundle yet, and we also haven't defined the zip bundle (see also #58).

oeway (approved):

FynnBe (approved):

constantinpape:
@esgomezm @frauzufall 
What do you think about the changes here? 
Is there still something you want to discuss in today's meeting or should we merge it before that?

README.md
```diff
@@ -55,19 +55,34 @@ Relative path to file with additional documentation in markdown.
 - `attachments`
 Dictionary of text keys and URI values to additional, relevant files.
 
+- [`packaged_by`]
+The person(s) that have packaged and uploaded this model. Only needs to be specified if different from `authors` in the root weights`, see `weights` for more details.
+
+- [`parent`] Parent model from which this model has been derived, e.g. by finetuning the weights of this model on a different dataset.
+  - `id` Url of another model available on bioimage.io or path to a local model in the bioimage.io specification. In addition to the
```
oeway, line 62: can we change `id` to `uri`.

constantinpape: Done

constantinpape:
@oeway @fjug @tomburke-rse @frauzufall @FynnBe @esgomezm 
I updated the PR to reflect the changes to `parent` and `data_type` / `data_range` we discussed today.
I think this PR should include the changes necessary to finalize 0.3.0 now.

Let me know if you approve / if there is still something that needs to change from your side.

oeway:
Ok, thanks! Could you please also take a look at https://github.com/bioimage-io/configuration/issues/60 ?

supported_formats_and_operations.md
```diff
@@ -61,6 +67,7 @@ Which consumer supports which preprocessing operation?
 |  `clip`                    | ?       | ?          | ?    | 
 |  `min_max`                 | ?       | ?          | ?    |
```
oeway, line 62: We should rename this to scale_min_max to be consistent with the post-processing and scale_linear.

Could you also check if we have the same definition for scale_min_max for pre- and post- processing? That way, we can simplify the implementation by sharing them.


BTW, do we really need to group them into pre- and post- processing for the implementation? Some functions might not make sense to be used in both, but some does. It might be easier to describe it as a pool of functions that one can used both for pre- and post-processing.

supported_formats_and_operations.md
```diff
@@ -90,12 +97,7 @@ The supported postprocessing operations.
     - `max` (if `mode == fixed`) the fixed max value
     - `reference_input` name of the input tensor to use as reference
   - `reference_implementation`
-- `scale_linear` scale the tensor with a fixed multiplicative and additive factor
-  - `kwargs`
-    - `gain` multiplicative factor
-    - `offset` additive factor
-    - `axes` the subset of axes to scale jointly. For example `xy` to scale the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
-  - `reference_implementation`
+- `scale_linear` scale the tensor with a fixed multiplicative and additive factor, sae as in [preprocessing](https://github.com/bioimage-io/configuration/blob/master/supported_formats_and_operations.md#preprocessing).
```
oeway, line 100: Typo? sae

supported_formats_and_operations.md
constantinpape: > We should rename this to scale_min_max to be consistent with the post-processing and scale_linear.

I agree,

> BTW, do we really need to group them into pre- and post- processing for the implementation? Some functions might not make sense to be used in both, but some does. It might be easier to describe it as a pool of functions that one can used both for pre- and post-processing.

Yes, I have also been thinking about this. The only thing that is specific to post-processing right now is the `reference_input`. But I agree that it would make it easier to just maintain a single list and then just write a comment for functions that are not suitable for pre/post-procesing.

I will change it later.

README.md
```diff
@@ -55,6 +55,13 @@ Relative path to file with additional documentation in markdown.
 - `attachments`
 Dictionary of text keys and URI values to additional, relevant files.
 
+- [`pacakged_by`]
+The person(s) that have packaged and uploaded this model. Only needs to be specified if different from `authors` in the root weights`, see `weights` for more details.
+
+- [`parent`] Parent model from which this model has been derived, e.g. by finetuning the weights of this model on a different dataset.
```
esgomezm, line 61: As there is `parent` here and in weights, could it be more specific? For example:

```yaml
- [`parent`] Parent model from which the trained weights of this model have been derived, e.g. by finetuning the weights of this model on a different dataset. For format changes of the same trained model checkpoint, see `models`.
```

supported_formats_and_operations.md
```diff
@@ -43,6 +43,12 @@ The supported preprocessing operations.
     - `min_percentile` the lower percentile used for normalization, in range 0 to 100.
     - `max_percentile` the upper percentile used for normalization, in range 0 to 100. Has to be bigger than `upper_percentile`.
   - `reference_implementaion`
+- `scale_linear` scale the tensor with a fixed multiplicative and additive factor
+  - `kwargs`
+    - `gain` multiplicative factor
+    - `offset` additive factor
+    - `axes` the subset of axes to scale jointly. For example `xy` to scale the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
```
FynnBe, line 46:
```suggestion
```
linear scale always applies to all axes

supported_formats_and_operations.md
```diff
@@ -43,6 +43,12 @@ The supported preprocessing operations.
     - `min_percentile` the lower percentile used for normalization, in range 0 to 100.
     - `max_percentile` the upper percentile used for normalization, in range 0 to 100. Has to be bigger than `upper_percentile`.
```
FynnBe, line 40:
```suggestion
    - `max_percentile` the upper percentile used for normalization, in range (1, 100]. Has to be bigger than `upper_percentile`. Note: We do not allow values <=1 to avoid mistakenly specifying the percentiles in [0,1]
```

supported_formats_and_operations.md
```diff
@@ -32,17 +32,23 @@ The supported preprocessing operations.
...
   - `reference_implementaion`
+- `scale_linear` scale the tensor with a fixed multiplicative and additive factor
+  - `kwargs`
+    - `gain` multiplicative factor
+    - `offset` additive factor
+    - `axes` the subset of axes to scale jointly. For example `xy` to scale the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
+  - `reference_implementation`
+- `scale_min_max` normalize the tensor to range 0, 1
+  - `kwargs`
+    - `axes` the subset of axes to normalize jointly. For example `xy` to normalize the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
```
FynnBe, line 50:
```suggestion
    - `mode` can be one of `per_sample` (min and max are computed for each sample individually), `per_dataset` (min and max are computed for the entire dataset)
    - `axes` the subset of axes to normalize jointly. For example `xy` to normalize the two image axes for 2d data jointly. The batch axis (`b`) is not valid here.
```

FynnBe:
Should we add a default `eps` value for zero_mean_unit_variance (and future eps kwargs elsewhere?).  0.0 or smth like 1.0e-6?
I think a value >0 would help to get closer outputs across implementations...

supported_formats_and_operations.md
FynnBe: Or we drop `scale_min_max` and leave `percentile` and `scale_linear`. This might help some backends to implement only one while complying with the config and keeps functionality nicely separated. What do you think?

constantinpape: I also looked at this again and I am not so sure what to do here. Right now combining everything into just one category doesn't really work because of different function signatures in pre/post-processing, e.g. for `scale_min_max`.

supported_formats_and_operations.md
constantinpape: see, I will fix it.

constantinpape: ah no, it's same :D

README.md
constantinpape: Thanks, I added your suggestion.

supported_formats_and_operations.md
FynnBe: I started https://github.com/bioimage-io/configuration/pull/61 to decouple this discussion from the rest of the changes in this PR. There I attempted to unify pre and postprocessing more..

fjug (commented):
Sorry, really late... my bad...
As far as I can see, neither Tom or myself see any fundamental problem.
(But we are sure that @frauzufall might disagree in January... and that is likely good!)

supported_formats_and_operations.md
constantinpape: Why?

constantinpape: I dpn't see any reason why we wouldn't allow to scale axes separately. In this case `gain` and `offset` need to be lists of course.

FynnBe: ah, of course, sorry for the confusion

constantinpape:
Ok, I think we resolved this and I am merging this :tada: !
We should be ready for 0.3.0 now, I will clean up issues and make a release later.

PR #66, opened 2021-03-16
Attachment field duplicated

esgomezm:
There are two different descriptions for `attachments`. 
Which one should we use?


constantinpape:
Good catch, but I think @oeway just fixed it on master, so we can close this.

oeway:
Sorry that I accidentally pushed, should make a PR.

PR #69, opened 2021-04-09
add dimension characters i,d and f

FynnBe:
@esgomezm encountered an example with the need to add a dimension to include multiple objects in one sample.
This lead to further discussion of the differences between image channels, categories, and computed features.

In today's meeting the remaining participant's @esgomezm , @oeway and me agreed on the addition of i (index/element), d (category/class) and f (feature) to cover the aforementioned cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ Name of this model. The model name should be human readble and only contain lett
 Timestamp of the initial creation of this model in [ISO 8601](#https://en.wikipedia.org/wiki/ISO_8601) format.
 
 - `description`
-A string containing a brief description. 
+A string containing a brief description.
 
 - `authors`
-A list of author strings. 
+A list of author strings.
 A string can be seperated by `;` in order to identify multiple handles per author.
 The authors are the creators of the specifications and the primary points of contact.
 
@@ -66,10 +66,10 @@ Relative path to file with additional documentation in markdown.
 The person(s) that have packaged and uploaded this model. Only needs to be specified if different from `authors` in the root weights`, see `weights` for more details.
 
 - `parent` \[optional\] Parent model from which the trained weights of this model have been derived, e.g. by finetuning the weights of this model on a different dataset. For format changes of the same trained model checkpoint, see `weights`.
-  - `uri` Url of another model available on bioimage.io or path to a local model in the bioimage.io specification. If it is a url, it needs to be a github url linking to the page containing the model (NOT the raw file). 
+  - `uri` Url of another model available on bioimage.io or path to a local model in the bioimage.io specification. If it is a url, it needs to be a github url linking to the page containing the model (NOT the raw file).
   - `sha256` hash of the weights of the parent model.
 
-- `inputs` 
+- `inputs`
 Describes the input tensors expected by this model.
 Must be a list of *tensor specification keys*.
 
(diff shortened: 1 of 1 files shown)


constantinpape:
Came up in our group meeting today; @akreshuk, @emilmelnikov  and me had some reservations.
The problem is that this proposal allows input data to be up to 9d data:
given the current proposal (if we understand this correctly) the input `[b, t, c, i, d, f, z, y, x]` is allowed.
However, all networks that we currently support, and will support in the foreseeable future, will allow for maximally 5d input.
So there is a lot of potential for headache caused by mapping the `axes` to the actual input internally.

Furthermore, the distinctions between `c`, `i`, `d` and `f` sound purely semantic to me. Maybe we should rather add this through axes metadata, for example with a new (optional) field `axes_description`.

On a related note: there is a discussion about the axes metadata happening for the ome ngff format as well:
https://github.com/ome/ngff/issues/35

oeway:
I think it's really depends on how we define the axis here. 

If we only allow 5 type of axes, say `bxyzc`, then we will have to choose from these 5 type of axes for every network. Now, say if we want a time axis, we will have to use perhaps `b` to represent while `t` is more intuitive. If we also allow `t`, then does it mean we allow 6 axes? This is to say, I wouldn't simply add the allowed axis types and take that as our allowed number of axes.

> Furthermore, the distinctions between c, i, d and f sound purely semantic to me. Maybe we should rather add this through axes metadata, for example with a new (optional) field axes_description.

As to my understanding, our current `axes` are semantic names of the axes, I mean otherwise you would simply use axes numbers. For me, the proposed `axes_description` is the same as our `axes` here. It provides semantical meaning or hint for each axis, a consumer software won't care about it if they only want to run the model and produce a tensor-- the `axes` definition only becomes useful when we want to render or display the tensor to the UI. For different cases, would be nice if we can choose from a rich set of vocabulary to express the semantical meaning.

tomburke-rse:
> Came up in our group meeting today; @akreshuk, @emilmelnikov and me had some reservations.
> The problem is that this proposal allows input data to be up to 9d data:
> given the current proposal (if we understand this correctly) the input `[b, t, c, i, d, f, z, y, x]` is allowed.
> However, all networks that we currently support, and will support in the foreseeable future, will allow for maximally 5d input.
> So there is a lot of potential for headache caused by mapping the `axes` to the actual input internally.
> 
> Furthermore, the distinctions between `c`, `i`, `d` and `f` sound purely semantic to me. Maybe we should rather add this through axes metadata, for example with a new (optional) field `axes_description`.
> 
> On a related note: there is a discussion about the axes metadata happening for the ome ngff format as well:
> [ome/ngff#35](https://github.com/ome/ngff/issues/35)

After fridays discussion and reading through this thread again, I've come to a conclusion I could live with (and support it ^^)
But first, let me reiterate my understanding so far and correct me if I'm wrong.
- XYZ have been declared as image axes
- T is used as time or channel in different models to provde a 4D aspect to images
- B is usually used for batches and I know that we use it in N2V and DenoiSeg

None of these have been officially documented and constrained as such. But why not do just that? We do want to visualize the data as an image or a table anyway and we also want to provide as much cross-plattform-compatibility as possible.
If we then add other axes and make it clear by documentation what they are used for (for example the ones Fynn mentioned) then we can all support them in our respective software and represent the data as tables and/or images, depending on axes used.

tl;dr: if we want to add more axes (and I see the need for that), we should constrain them and only **support** them.

FynnBe:
> * T is used as time or channel in different models to provde a 4D aspect to images

so far we have `t` for time and `c` for channels.

FynnBe:
Some general thoughts aimed at searching for common ground and compromises everyone can live with:

1. What do axes letters (like those we currently use) actually mean?
- `tzyx`: time and space are 'clean' physical dimensions, with zyx refering to a 3d (righthanded?) space.
- `b`: special, but straight forward batch dimension: really just a list of data samples that works as a tensor dimension, because we generally optimize our code to do so for efficient parallel computing. (There are exceptions though, different samples do not have to have the same shape, which opens the next question: Do we want/need to coer this case (in the future)?)
- `c`: opinions start to diverge. 
Either this 'other' dimension encodes values for the same sample/time/location. Then `i`, `d` and `f` just add semantic meaning to `c`. From this point of view `c` just happens to encode image channels in many use cases, but might as well be anything else, potentially flattened into this one 'rest' dimension. 
Or `c` is associated with 'image channel' and differentiating to other semantic meanings justifies going (at least conceptually) to a higher dimensions.

2. What description of input/output (axes) do we need? (This part starts to be colored by my own opinion, I'm afraid)
- The input/output description needs to encode the nature of the data to the extent relevant to apply the model and allow consumer software to show data appropriately.
- Only if the input/output can be interpreted in code -- not by a human -- we can provide a smooth end user experience. Thus we need to balance **easy of use VS flexibility** (keeping it simple, while we don't want to exclude any model devs or consumer software from the model zoo, or make them escape to the `config` section by default).
- Beyond 'understanding' model input/output it's up to the consumer software to display this to the end user. High flexibility leads to a need for high configurability. Complex data requires the end user to choose how to look at the data, while simple data can be shown in a fixed, unconfigurable manner.

My conclusions so far:
- We do not only aim at a framework to describe arbitrary data (like the ngff format), but would like to cater (or even limit) the data to the capabilities of our consumer software.
- While our consumer software might include amazingly specific applications (in the future), we have to pay special attention to the more common use cases with the common tools and common models (tends to be on the simpler-ish side, but already includes tables).
 - While we can mostly get away with limiting i/o data to a certain dimension limit (~6-8), we do have to account for different types of dimensions (-> image vs table, etc.). Also limiting the number of dimensions will exclude some models and not make them comply with lower dimensions instead.
- We should not forget that we want to promote consumer cross-compatibility, which complex data descriptions hamper.

Proposal:
- We do add a few (pseudo) dimensions (`i`, `f`, `d`) to allow for straight forward description of almost all i/o data, which serves as a more detailed visualization hint (this list of `c` like axes may be extended in the future, but we will try to keep it short). 
- We agree not to duplicate any axis letter (chances are those dimensions can very well be flattened, which makes the model more accessible to a wider range of consumer software). 
- We emphasize the close relationship of channel like (pseudo) dimensions in the documentation to encourage model devs to limit their use to one of them, which allows to map those to the `c` dimension (e.g. in software that can 'only' handle 6d data) and 'keep it simple', while allowing higher dimensional models (max 9d)
- As this turned out more complex than initially assumed, we add a key `axes_description` with a list of human readable short (single line + char limit) descriptions (maybe nested for 'columns'), that consumer software could potentially display. (alo: maybe fix description for `b`). Details are left for further discussion.
- Left out for further discussion: Add keys like  `axes_unit`/`axes_resolution` to further clarify the nature of the i/o data.

example draft (only for a scalar + table, as we all can image "simple" images/videos in this format):
```yaml
inputs:
- name: screenshot_of_source_code
  axes: bxy
  ...

# example with fixed columns/ small table
outputs: 
  - name: bug_density  # scalar per image
    axes: b
    ...
  # table with individual bugs
  - name: bug_size  # first column
    axes: bf
    shape: [1, 100]    
    ...
  - name: bug_age  # second column
    axes: bf
    shape: [1, 100] 
    ...

# example with big table 
outputs:
  - name: bug_density  # scalar per image
    axes: b
    ...
  - name: bug_table    # table with individual bugs
    axes: bif  
    shape: [1, 100, 2]  # btw, we don't allow for an unknown size (yet), which would be important for tables, there are not 100 bugs on every page of code after all.
    axes_description: ["batch", "bug", ["size", "age"]]   # idea: nested axes description (has to match shape). Details of this go a bit beyond the scope of this issue/PR for now.
```

note that in the example axes `f` has size 100 in the small table example, but 2 in the big table example...  This minimal example demonstrates how much variability these limited options already give to developers to define their model output in ambiguous ways.

FynnBe:
> Proposal:
> 
> * We do add a few (pseudo) dimensions (`i`, `f`, `d`) to allow for straight forward description of almost all i/o data, which serves as a more detailed visualization hint (this list of `c` like axes may be extended in the future, but we will try to keep it short).
> * We agree not to duplicate any axis letter (chances are those dimensions can very well be flattened, which makes the model more accessible to a wider range of consumer software).
> * We emphasize the close relationship of channel like (pseudo) dimensions in the documentation to encourage model devs to limit their use to one of them, which allows to map those to the `c` dimension (e.g. in software that can 'only' handle 6d data) and 'keep it simple', while allowing higher dimensional models (max 9d)
> * As this turned out more complex than initially assumed, we add a key `axes_description` with a list of human readable short (single line + char limit) descriptions (maybe nested for 'columns'), that consumer software could potentially display. (alo: maybe fix description for `b`). Details are left for further discussion.
> * Left out for further discussion: Add keys like  `axes_unit`/`axes_resolution` to further clarify the nature of the i/o data.

/polls "I am OK with this proposal" "I might be OK with this, if minor changes that I briefly commented on below are addressed" "This is going nowhere! We have to take a completely different angle, I'll make a separate issue to lay out my plans"

constantinpape:
I like adding `axes_description` and I think we can use it to express the custom metadata e.g. for informing a viewer how a channel should be displayed. I think this way we don't need to add new letters like `i`, `d`, `f` here. See discussion in #70.

constantinpape:
I think we have a very good intermediate solution (https://github.com/bioimage-io/spec-bioimage-io/issues/70#issuecomment-824036922). I think it's probably best to close this PR and open a new one for it.
@FynnBe do you want to give it a go?

constantinpape:
We decided to first do an intermediate change where we introduce just `i`, see #75.
And then we will go for a change in the `axes` layout, see https://github.com/bioimage-io/spec-bioimage-io/issues/70#issuecomment-825737433. Hence, this PR is deprecated and I am closing it.

PR #71, opened 2021-04-22
move halo out of output shape

k-dominik:
seems to be misplaced (at least the validator will mark it as an error in such a spec)


oeway (approved):
LGTM

PR #72, opened 2021-04-23
fix incorrect reference to `upper_percentile`

k-dominik:
I suspect there is a typo in the `scale_range` description.


oeway (approved):
Thanks!

PR #75, opened 2021-04-23
Add axis name i

constantinpape:
In preparation of the updated axes defintion, we introduce the `i` axis label and bump the PATCH version.

@esgomezm this should enable expressing the deepImageJ models that predict tables in the spec without making the change to 0.4.0. Let me know if you see anything that should be changed in this PR.


oeway:
I added the table for axes description.

PR #79, opened 2021-05-28
Use SPDX license identifier

oeway:
Right now, we haven't define how the license identifier should be specified. I saw Zenodo uses SPDX identifier and I think we should basically enforce the same. 

Here is a list of them: https://spdx.org/licenses/

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ It is recommended to:
  * the model folder can be packaged into a zip file, and the package name should use a descriptive name + `.model.zip`.
  * use or upgrade to the latest format version
 
-## Current `format_version`: 0.3.1
+## Current `format_version`: 0.3.2
 
 A model entry in the bioimage.io model zoo is defined by a configuration file `model.yaml`.
 The configuration file must contain the following fields; optional fields are followed by \[optional\].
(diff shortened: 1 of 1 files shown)


FynnBe:
I'd be in favor to limit valid licenses (to the SPDX list), too 👍

FynnBe:
looking at the python package @oeway proposed in https://github.com/bioimage-io/python-bioimage-io/issues/68 I have some follow up questions, that should be be discussed here instead:

This packages provides additional information on each license, so should we:
> isDeprecatedLicenseId - bool value telling you whether this license definition is oudated or not.
- warn for a deprecated licence (or fail?)?
> isFsfLibre - bool value telling you whether this license is "free software" by the definition of FSF.
> isOsiApproved - bool value telling you whether this license is "open source" by the definition of OSI.
- require it to be free/open source?? (I think we agreed on not enforcing any specific license groups, but I want to make sure I didn't just make that up...)

oeway:
> This packages provides additional information on each license, so should we:
> 
> > isDeprecatedLicenseId - bool value telling you whether this license definition is oudated or not.
> 
> * warn for a deprecated licence (or fail?)?
> 
> > isFsfLibre - bool value telling you whether this license is "free software" by the definition of FSF.
> > isOsiApproved - bool value telling you whether this license is "open source" by the definition of OSI.

We can perhaps allow anything in the list, but produce deprecation note in the output?

> 
> * require it to be free/open source?? (I think we agreed on not enforcing any specific license groups, but I want to make sure I didn't just make that up...)

I think we don't need to enforce any further rule as a start, the current spec only defines a model, which does not necessary need the source code (e.g. one can make a pytorch model in torch script format without providing the source code). I mean open source licence are typically apply to software, but here, for model, notebook, and dataset, those "content-oriented" license such as CC-BY is perhaps more relevant.

One way to avoid super long discussion is to perhaps leave it open-ended for now ;)

constantinpape (approved):
Very good idea @oeway.

FynnBe:

FynnBe:
implemented in 5599987

PR #80, opened 2021-05-28
Move spec and integrate other changes (format_version: 0.3.2)

FynnBe:
Move over spec definition, validation and generated documentation from https://github.com/bioimage-io/python-bioimage-io

PR #81, opened 2021-05-28
Support describing author affiliation

oeway:
Currently, the authors in the `authors` list is author name string only. But Zenodo requires each author has `name` and `affiliation`. I think it might be better if we can also support author dictionary.

There can be two ways to support this, 1) allowing both a string and a dictionary so we can be backward compatible 2) migrate to dictionary only, and we can provide automatic conversion in the packager (similar to the `axes`).

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ It is recommended to:
  * the model folder can be packaged into a zip file, and the package name should use a descriptive name + `.model.zip`.
  * use or upgrade to the latest format version
 
-## Current `format_version`: 0.3.2
+## Current `format_version`: 0.3.3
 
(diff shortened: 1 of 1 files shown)


constantinpape:
> There can be two ways to support this, 1) allowing both a string and a dictionary so we can be backward compatible 2) migrate to dictionary only, and we can provide automatic conversion in the packager (similar to the `axes`).

I would prefer option number 2 to keep the spec simpler. But I don't have a strong opinion here, so I am fine with either solution.

FynnBe:
I'd also prefer option 2. We can add an automatic conversion for the current author lists

FynnBe:

FynnBe:
implemented in d63f019d6c94b2821cec139558e359463bef54e8 (part of #80)

PR #82, opened 2021-05-28
Add id key for storing the doi

oeway:
We need an `id` key to store the generated `doi` for the model. Note that we already supported `id` key in the minimal RDF:
```
"id": {
            "type": "string",
            "description": "unique identifier for the resource, you can also use a `doi` e.g. generated from [zenodo](https://zenodo.org/)"
},
```

So it's natural to have it also here as an optional key.


constantinpape (approved):

FynnBe:

FynnBe:
implemented in 7e9effe

PR #83, opened 2021-05-29
Add a `type` and `version` key

oeway:
In the RDF definition we have the `type` and `version` key, it would be nice if we can add it for the models as well.

Todo: bump version and add changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,9 @@ If a field is followed by \[optional\]\*, they are optional depending on another
 Version of this bioimage.io configuration specification. This is mandatory, and important for the consumer software to verify before parsing the fields.
 The recommended behavior for the implementation is to keep backward compatibility, and throw error if the model yaml is in an unsupported format version.
 
+- `type`
+The value must be `model`, used to differentiate with other RDF types such as `dataset`, `application`.
+
 - `name`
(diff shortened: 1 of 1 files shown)


constantinpape (approved):

constantinpape:
@oeway I think all the changes in #79, #81, #82 and #83 make a lot of sense. But I suggest we introduce all of them in a single version bum from 0.3.1 to 0.3.2. That would make it easier to make actual releases on gh for these versions.

oeway:
> @oeway I think all the changes in #79, #81, #82 and #83 make a lot of sense. But I suggest we introduce all of them in a single version bum from 0.3.1 to 0.3.2. That would make it easier to make actual releases on gh for these versions.

Agreed, I am thinking the same while I add more, will make sure I merge them.

FynnBe:

FynnBe:
should the version be `None` if not present or should it be 0.1.0 (my favorite, so it's never None and always comparable)?

oeway:
> should the version be `None` if not present or should it be 0.1.0 (my favorite, so it's never None and always comparable)?

No strong opinion here, and yes, I think we can default it to 0.1.0 if not present.

FynnBe:
implemented in https://github.com/bioimage-io/spec-bioimage-io/pull/80/commits/d03984dd14af0afe4b2bd1ad0e6f4252beaa6d90 (part of #80)

Issue #87, opened 2021-06-02
build_spec support for keras/tf


maweigert:
Hi, 

It appears that [`build_spec`](https://github.com/bioimage-io/python-bioimage-io/blob/7d47a8cdf311348d9ab2ceca13fd925f708c4218/bioimageio/spec/utils/build_spec.py#L215) currently does not support keras.tf `hdf5` weights, minimal example:

```python 
import bioimageio

spec = bioimageio.spec.utils.build_spec(
        model_kwargs={},
        weight_uri='https://github.com/JonathanCMitchell/mobilenet_v2_keras/releases/download/v1.1/mobilenet_v2_weights_tf_dim_ordering_tf_kernels_1.0_224_no_top.h5',
        weight_type='keras_hdf5',
        test_inputs=['https://github.com/bioimage-io/pytorch-bioimage-io/raw/master/specs/models/unet2d_nuclei_broad/test_input.npy'],
        test_outputs=['https://github.com/bioimage-io/pytorch-bioimage-io/raw/master/specs/models/unet2d_nuclei_broad/test_input.npy'],
        name='', description='', authors=[],tags=[], license='', documentation='', covers='', dependencies='',cite=[]
    )
```

```
TypeError: Parameters to generic types must be types. Got 0.
```

constantinpape:
Good point. I didn't add it yet because I didn't have an example at hand when working on this.
I will try to do this later and let you know once it's there.
@oeway @esgomezm , do you have a good example with the keras hdf5 format uploaded somewhere, that I could use for the tests here?

esgomezm:
Hi @constantinpape,

Do you need it in TF1 or TF2? I have some in TF1.15. Also, do you need the weights or the entire model?

constantinpape:
> Do you need it in TF1 or TF2? I have some in TF1.15. Also, do you need the weights or the entire model?

The TF version does not really matter, I just need some example model for the test.
Ideally, I would need a link to a valid model yaml hosted somewhere, like here: https://github.com/bioimage-io/python-bioimage-io/blob/7d47a8cdf311348d9ab2ceca13fd925f708c4218/tests/test_bioimageio/test_spec/test_utils/test_build_spec.py#L51-L52

esgomezm:
Ok, so this model has a keras model inside: https://github.com/deepimagej/models/blob/master/fru-net_sev_segmentation/model.yaml

Let's see if it works!

constantinpape:
[...]

PR #89, opened 2021-06-04
Add support for missing weight formats to build_spec

constantinpape:
See #87. 

@FynnBe I think the changes here are correct, however the tests fail. I think the reason is that the example model for keras/tf at https://github.com/deepimagej/models/blob/master/fru-net_sev_segmentation/model.yaml does not conform with the current spec.
It would be good to check why and see if this is an issue in the checker or with the model itself.


FynnBe:
changes here look fine to me, but the tests break (still after fixing issues in 'move_spec' branch) due to the model not being up to date...

Issue #91, opened 2021-06-04
Changes to 0.3.2


constantinpape:
As discussed in the meeting today.
- Rename `id` to `_id` and treat it as an internal field that should not be provided by users. This is then autogenerated by the website (e.g. by the zenodo upload script).
- make affiliation optional (default: null)
- change default version: null
- Loosen requirements for the `documentation`: we should allow the two following use cases:
    - Having a local .md file, in this case the documentation can be rendered in the bioimage.io card  
    - Having an url in some existing documentation, in this case, the model card should just show something like "The documentation for this model can be found at this link ...".
    
cc @FynnBe @oeway @esgomezm 

Edit: copied the two additional changes on default values from the meeting notes

constantinpape:
@oeway the part of the documentation came up after you left the meeting.
@esgomezm and I think that we should allow linking to external documentation pages in order to avoid duplicate work for models where documentation exists already.

oeway:
> As discussed in the meeting today.
> 
> * Rename `id` to `_id` and treat it as an internal field that should not be provided by users. This is then autogenerated by the website (e.g. by the zenodo upload script).

Good!

> * Loosen requirements for the `documentation`: we should allow the two following use cases:
>   
>   * Having a local .md file, in this case the documentation can be rendered in the bioimage.io card
>   * Having an url in some existing documentation, in this case, the model card should just show something like "The documentation for this model can be found at this link ...".

I would recommend that we enforce a local README file, for a couple of reasons:
 * As a website, we want to keep uses within the website while browsing different models. If each and every model provide a link, the user will need to switch tabs to read them. Which will cause bad experience.
 * External links can break, or from security point of view, we should limit the number of external links.
 * A local document is the best way to guarantee the consistency between the models and the docs. Otherwise, one might have their documentation linked to GitHub README which can be easily updated while the model is frozen.
 * For exported offline packages, it's nice to have a readme file within the package. And a consumer software can easily render a local documentation interface.
 * It is also recommended for many of the archive that to save a README file in the online deposit of dataset.
 * Forcing a local documentation will allow us to tap into the format later, and give better recommendations later for the docs, e.g. implement model cards idea

After all, you can still link to your external documentation within the local README file.

constantinpape:
> After all, you can still link to your external documentation within the local README file.

That is true, although it feels a bit unnecessary to enforce a local file only to provide a link.
And I don't really think that it's good practice to encourage people to duplicate documentation:
[...]

PR #97, opened 2021-06-05
Change 0.3.2

FynnBe:
changes as summarized in https://github.com/bioimage-io/spec-bioimage-io/issues/91#issuecomment-854905623

remaining problem: auto conversion for `documentation`:
- [ ] converting an invalid input (e.g. a url) to a local md file requires the __whole package__, not just the yaml content
    - not a problem for 'validate-spec' (<- should be renamed 'validate-model'.?), which takes a model.yaml path, but a problem for 'validate-bioimageio-manifest' (should be renamed 'validate-manifest'?)
 - [ ] requires python packager to create package from yaml url


FynnBe:
relates to #98

bioimageio/spec/__main__.py
```diff
@@ -14,19 +14,24 @@
 app = typer.Typer()  # https://typer.tiangolo.com/
 
 
+@app.command()
+def package_model(model_yaml: Path, auto_convert: bool = False):
```
constantinpape, line 18: General comment: Can we add docstrings to the CLI? I think it would be good to get something meaningful on `package_model --help` etc.
(Fine to do in a separate PR, just wanted to point this out.)

bioimageio/spec/utils/maybe_convert_to_v0_3.py
```diff
@@ -154,11 +153,19 @@ def convert_model_v0_3_1_to_v0_3_2(data: Dict[str, Any]) -> Dict[str, Any]:
...
+        if authors_update is not None:
+            for a, u in zip(weights_entry["authors"], authors_update):
+                a.update(u)
+
+    # documentation (if not a local relative md file, pretend content is in local 'auto_converted_documentation.md')
+    class DocSchema(Schema):
+        doc = schema.Model.documentation
+
+    if DocSchema().validate({"doc": data["documentation"]}):
+        data["config"] = data.get("config", {})  # make sure config exists
```
constantinpape, line 165: Why is "config" involved here? I'd assume that we just put the content of the "documentation" field into `AUTO_CONVERTED_DOCUMENTATION_FILE_NAME`?

constantinpape:
> changes as summarized in [#91 (comment)](https://github.com/bioimage-io/spec-bioimage-io/issues/91#issuecomment-854905623)
> 
> remaining problem: auto conversion for `documentation`:
> * converting an invalid input (e.g. a url) to a local md file requires the **whole package**, not just the yaml content
>   
>   * not a problem for 'validate-spec' (<- should be renamed 'validate-model'.?), which takes a model.yaml path, but a problem for 'validate-bioimageio-manifest' (should be renamed 'validate-manifest'?)
> 

If I understand this right, the problem is that if we only have the yaml we cannot fully auto convert with missing documentation.
But if we only have the yaml we cannot fully validate the package in any case, because, for example, we can't check if the .md file exists.
So shouldn't it be fine to just change the documentation field and pretend that the auto converted .md file would exist?

bioimageio/spec/utils/maybe_convert_to_v0_3.py
FynnBe: that would not pass the schema, so I thought `config` would be a good place too keep it.

bioimageio/spec/__main__.py
FynnBe: totally! 👍

FynnBe:
> So shouldn't it be fine to just change the documentation field and pretend that the auto converted .md file would exist?

yes, that's what this PR does so far...

bioimageio/spec/utils/maybe_convert_to_v0_3.py
constantinpape: Yes, but why are you adding 
```python
data["config"][AUTO_CONVERTED_DOCUMENTATION_FILE_NAME] = data["documentation"]
```
in the first place?
Is it just to keep track of the fact that the documentation has been auto generated?

FynnBe: to save the original content of `data["documentation"]` (that would need to be written to the AUTO_CONVERTED_DOCUMENTATION_FILE_NAME file)

FynnBe: as the converter only has the yaml content as dict it cannot write the file as it wouldn't know where to.

constantinpape: Ok, makes sense. Could you sum this up in a comment though? This is not obvious from the code alone.

constantinpape:
@FynnBe should we go ahead and merge this, I think all the changes are approved now and the test issues are only due to https://github.com/deepimagej/models/pull/22.

FynnBe:
the following todos can be left for future PRs:
- [ ] reactive ci for tests (potentially with #103)
- [ ] clean up `schema.Model` base class `schema.RDF` (mostly just move around fields)
- [ ] CLI to verify models and manifest for specific versions (and not juts the latest)
- [ ]  add more example specs here?

oeway:
Would it make more sense to make it `0.4.0` instead?

FynnBe:
> Would it make more sense to make it 0.4.0 instead?

the stricter requirements for `license` are the only thing that could break a 0.3.1 model. It would be the cleanest to go 0.4.0, but I think it's not worth the effort of changing this now, as that part is already in master.

if this really is an issue I would change it to only warn about the license, but we do not have an established "warn about deprecattion" test/validator/tool, which we should add for better transitions to future versions

constantinpape:
Yeah, I think we should have been more careful when doing the initial bump to 0.3.2 and named it 0.4.0 already.
I think the changes here are just to fix some of the inconsistencies we had when bumping to 0.3.2 so we should merge it without complicating it further.

constantinpape (approved):

constantinpape:
Let's also move the  checklist from https://github.com/bioimage-io/spec-bioimage-io/pull/97#issuecomment-859533766 to some issue once this is merged so that we can keep track of it.

FynnBe:
> Let's also move the checklist from #97 (comment) to some issue once this is merged so that we can keep track of it.

created #105

PR #99, opened 2021-06-05
enable working with (older) spec versions

FynnBe:
restructures the code to improve validation, handling and conversion of and between older spec versions


tests/test_build_spec.py
```diff
@@ -1,12 +1,12 @@
 import os
 from pathlib import Path
+
 from bioimageio.spec import maybe_convert_model, schema
-from bioimageio.spec.utils import yaml
+from bioimageio.spec.common_utils import yaml
+from bioimageio.spec.latest.build_spec import _get_local_path, build_spec
```
constantinpape, line 6: This is just a minor point, but it's good practice to import the functionality being tested in the test instead of importing it globally, so that potential ImportErrors are caught on an individual test level.

FynnBe:
travis tests are not runnig, switching to github actions in
tests pass locally (with expected exception of 3 tests based on https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml)

@constantinpape  this is ready to be merged (into branch change_0.3.2)

for now `bioimageio.spec verify-sepc` and `... verify_bioimageio_manifest` will are still trying to validate (with or without auto conversion) according to newest version (0.3.2). I leave it for a future PR for the CLI tools to add a paramter to control validating the specified version instead.

constantinpape (approved):
Looks good to me. Go ahead and merge when you think this is ready.

PR #102, opened 2021-06-10
Move manifest and test_mainfest from python-bioimage.io

constantinpape:
I think this was lost when moving things from python-bioimage-io earlier.
I am not quite sure why the manifest is empty, cf. https://github.com/bioimage-io/python-bioimage-io/pull/73.


FynnBe:
there are no models here that should be shared and find there way to bioimage.io (the rf model is based on "pickle weights" which are not officially supported and only for testing and developing)
thinking about this "pickle" is probably listed as valid weight format though... (it really shouldn't!)

tests/test_manifest.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+import yaml
+
+from bioimageio.spec import load_spec, utils
+
+MANIFEST_PATH = Path(__file__).parent / "../manifest.yaml"
+
+
+def pytest_generate_tests(metafunc):
```
FynnBe, line 11: this generation of individual model tests from the manifest is outdated as we now have the `bioimageio.spec verify-bioimageio-manifest` command.

This test_manifest.py really does not test the manifest, but its content

constantinpape:
Thanks for the clarification, @FynnBe.
I think we can then just close this PR

> thinking about this "pickle" is probably listed as valid weight format though... (it really shouldn't!)

Yes, it's listed as an official format. Maybe we should just remove the whole pickle thing? I think having it is confusing us more than helping. I will make a separate issue about this.

PR #103, opened 2021-06-11
add tox gh workflow

FynnBe:


constantinpape (approved):

FynnBe:
unfortunately it does not seem like the tests are actually being executed atm... @k-dominik is looking into this more

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox tox-gh-actions
+    - name: Test with tox
+      run: tox
```
k-dominik, line 27: I don't have push access here I guess - so you have to debug it yourself.

how about:

```suggestion
      run: tox -vvv
```

for some more insights into what's going on

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+
+    steps:
+    - uses: actions/checkout@v1
```
k-dominik, line 17:
```suggestion
    - uses: actions/checkout@v2
```

v2 seems to be the current one (since 2019)

constantinpape:
@k-dominik I invited you to the bioimageio org, I think then you should have all the necessary rights to make changes here.

k-dominik:
allright, now it seems to produce valid failures :)

looks like one has to configure everything redundantly - specifying python version 3 times... geez

constantinpape:
Ok, there are 2 errors left when evaluating the deepImageJ model.

First one: https://github.com/bioimage-io/spec-bioimage-io/pull/103/checks?check_run_id=2804708560#step:5:172
```
 E           marshmallow.exceptions.ValidationError: {'_schema': ["invalid 'tensorflow_version' entry for weights format keras_hdf5"]}
```
This is pretty straightforward: `keras_hdf5` does not support the field `tensorflow_version`. I would assume that this is a mistake in the spec, because it makes sense that one would need it for a keras model.
@esgomezm is this correct?

Second one: https://github.com/bioimage-io/spec-bioimage-io/pull/103/checks?check_run_id=2804708560#step:5:316
```
   E           marshmallow.exceptions.ValidationError: {'weights': defaultdict(<class 'dict'>, {'tensorflow_saved_modle_bundle': {'key': ['Must be one of: pickle, pytorch_state_dict, pytorch_script, keras_hdf5, tensorflow_js, tensorflow_saved_model_bundle, onnx.']}})}
```
This one is bizarre: the validator complains about the weight format `tensorflow_saved_modle_bundle` (emphasis on the **modle**). However, it is correctly spelled **model** in the source yaml, see https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml.
I have no clue what that's about.

FynnBe:
> Second one: https://github.com/bioimage-io/spec-bioimage-io/pull/103/checks?check_run_id=2804708560#step:5:316
> 
> ```
>    E           marshmallow.exceptions.ValidationError: {'weights': defaultdict(<class 'dict'>, {'tensorflow_saved_modle_bundle': {'key': ['Must be one of: pickle, pytorch_state_dict, pytorch_script, keras_hdf5, tensorflow_js, tensorflow_saved_model_bundle, onnx.']}})}
> ```
> 
> This one is bizarre: the validator complains about the weight format `tensorflow_saved_modle_bundle` (emphasis on the **modle**). However, it is correctly spelled **model** in the source yaml, see https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml.
> I have no clue what that's about.

I found 'tensorflow_saved_modle_bundle' in build_spec.... https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml is used in test_build_spec...

constantinpape:
> I found 'tensorflow_saved_modle_bundle' in build_spec.

Ah ok, so the typo was in another place then I expected. I am not sure how this ended up being thrown as a marshmallow error, but nvm. Thanks for catching this. 

I guess what's left to do is adding `tensorflow_version` to the `keras_hdf5` weight format.

FynnBe:
> I guess what's left to do is adding tensorflow_version to the keras_hdf5 weight format.

done in https://github.com/bioimage-io/spec-bioimage-io/pull/103/commits/e147c51836a49109a57450320ccdb875ad80f42e

looking at that part I am wondering if we should require `tensorflow_version` for the tf derived weights (["keras_hdf5", "tensorflow_js", "tensorflow_saved_model_bundle"]) instead of just allowing it.

Same for `opset_version` and "onnx"

constantinpape:
> looking at that part I am wondering if we should require `tensorflow_version` for the tf derived weights (["keras_hdf5", "tensorflow_js", "tensorflow_saved_model_bundle"]) instead of just allowing it.

I think this is something we have to discuss separately with @esgomezm. For example the [example model](https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml) does not have a tf version for "tensroflow_js".

constantinpape:
Also, I think we can't support python 3.6 due to missing type annotations:
```
   E     File "/home/runner/work/spec-bioimage-io/spec-bioimage-io/bioimageio/spec/shared/fields.py", line 2
  E       from __future__ import annotations
  E                                        ^
  E   SyntaxError: future feature annotations is not defined
```

FynnBe:
> I think this is something we have to discuss separately with @esgomezm. For example the example model does not have a tf version for "tensroflow_js".

yes. I added warnings for these cases now. Enforcing it would be a version bump anyhow...

FynnBe:
@constantinpape  did you cancel the tests or do you have any idea why they are getting canceled?

constantinpape:
> @constantinpape did you cancel the tests or do you have any idea why they are getting canceled?

They are getting canceled because the python 3.6 one fails and the "eary stopping" option is set to true (I don't recall the actual name of this parameter rn).

FynnBe:
> Also, I think we can't support python 3.6 due to missing type annotations:

I think we don't need the delayed evaluation of type annotations anymore... 
However, we depend on dataclasses. There is a backport for py3.6: https://pypi.org/project/dataclasses/ (also available on conda)....

k-dominik:
if we could agree to move this to conda (away from pip) we can do this transparently for python 3.6, by simply adding dataclasses to the reqs - conda will do the "smart" thing there and use the backport if necessary

k-dominik:
alternatively we could also require python >=3.7 :)

FynnBe:
https://github.com/bioimage-io/spec-bioimage-io/pull/103/commits/5b942433ddc48e71d440a0071ae3263c368c4233 fixes the dataclasses with pip, but apparently the backport has some different inner structure:
```
  bioimageio/spec/v0_3/raw_nodes.py:27: in <module>
      latest_version = FormatVersion.__args__[-1]  # todo: py 3.9: use typing.get_args
  E   AttributeError: '_Literal' object has no attribute '__args__'
```

we need a backport for `get_args`

constantinpape:
I think we can safely drop python 3.6.

FynnBe:
yes, somehow `get_args` is not in typing_extensions for python 3.6...  Let's drop it for now (hasn't ever been there anyway)

constantinpape:
>  Let's drop it for now (hasn't ever been there anyway)

Ok, will you do this @FynnBe?

constantinpape:
Hooray, the CI is green and I will merge it!

PR #115, opened 2021-06-14
Generate docs and schema with github actions

oeway:
This PR uses github actions to generate both the docs and schema, the generated file will be stored in the [gh-pages branch](https://github.com/bioimage-io/spec-bioimage-io/tree/gh-pages). 

## Why use CI to generate?
I think this should be a better approach compared to pre-commit hooks which rely on certain setup and won't work for edit made directly in github website. It is also nice to separate the true source code from the generated files into different branches.


oeway:
This PR is ready for preview.

FynnBe (approved):
very neat!

oeway:
@FynnBe thanks for the review, these is an issue for running the generating json spec, if I enable the command, the CI fails said no module named bioimageio.spec..., but the wired thing is that the generate docs command runs without issue. Any idea?

FynnBe:
> @FynnBe thanks for the review, these is an issue for running the generating json spec, if I enable the command, the CI fails said no module named bioimageio.spec..., but the wired thing is that the generate docs command runs without issue. Any idea?
> 
> Let's fix it in another PR

👍  I'll have a look at it soon

PR #116, opened 2021-06-14
Change `rdf.bioimage.io.yaml` to `rdf.yaml`

oeway:
The file name `rdf.bioimage.io.yaml` is a bit too long, and since `rdf.yaml` is not that common anyway, I think it's better to just use the short version.


FynnBe (approved):
👍

Issue #118, opened 2021-06-15
Serialization of raw_nodes.Model writes out all missing fields with some default


constantinpape:
Serializing a `raw_nodes.Model` writes out the default values for all fields, also the ones not passed to the `raw_nodes.Model` constructor.
```python
import bioimageio.spec as spec
model = spec.raw_nodes.Model(...)  # create the model with some missing args

# round trip for validation
ser = spec.schema.Model().dump(model)
model = spec.schema.Model().load(ser)
ser = spec.schema.Model().dump(model)

# write to yaml
with open("model.yaml", "w") as f:
  spec.utils.yaml.dump(ser, f)
```
The resulting "model.yaml" will contain all fields with defaults (mostly `null` or empty lists). They should not be written at all.

FynnBe:
After loading a model from a yaml file optional fields are filled with `None` or another appropriate 'empty' value like empty list or dict.
The reasoning is that users of the bioimageio.spec parser and validation should not need to know about the special 'missing' value and that certain missing fields are equivalent to those default values, thus your produced "model.yaml" is exactly equivalent after loading it again. 
This is desired behavor to make it easier to work with the parsed model and acknowledges that there is no logic difference between writing a model.yaml with e.g.
```yaml
...
cite: []
```
or leaving out the `cite` field altogether.

The choice of 'missing' values in the raw nodes allows us to specify meaningful default values only in one place (the schema), thus leaving out an optional field in raw nodes or in a model.yaml always leads to the same default value. This is why creating raw_nodes 'manually' is somewhat of an illegal move and needs a round trip to set the correct defaults.

constantinpape:
Ok, I see. This makes using the raw_nodes very awkward, because they are different after validation / serialization and the state that comes back is very "non-intuitive", e.g. all weights have an entry for "opset version" although this only makes sense for onnx.
This also prevents using the library for things like "read the model yaml, add or update some fields and serialize it again".

Anyway, I can work around this for by writing a function that just removes all empty fields before serialization.

constantinpape:
I have added a workaround in #119

FynnBe:
let's use this space to dicsuss possible improvements or if we want to keep it this way (and find another approach without workarounds)...

FynnBe:
>  e.g. all weights have an entry for "opset version" although this only makes sense for onnx.

This can be changed by adding specific WeightEntry raw_nodes/schema.

[...]

PR #119, opened 2021-06-15
Add utility function for adding additional weight formats to model spec

constantinpape:


constantinpape (commented):
@FynnBe This adds another utility functionality to add weight formats to an existing model, and a workaround caused by the inconsistent serialization of raw_nodes, see #118.
In addition, the current code does not allow "model.yaml" as valid name, however this is expected by the zenodo uploader from @oeway. It would be nice to converge on something that works for the uploader.

The test failures are due to the fact that model yaml names are currently not checked.

bioimageio/spec/v0_3/utils.py
```diff
@@ -35,9 +35,9 @@ def _(source: dict, root_path: Optional[pathlib.Path] = None) -> Tuple[raw_nodes
 def _(source: os.PathLike, root_path: Optional[pathlib.Path] = None) -> Tuple[raw_nodes.Model, pathlib.Path]:
     source = pathlib.Path(source)
 
-    suffixes = source.suffixes
-    if len(suffixes) < 2 or suffixes[-1] not in (".yml", ".yaml") or source.suffixes[-2] != ".model":
-        raise ValidationError(f"invalid suffixes {''.join(suffixes)} for source {source}")
+    # suffixes = source.suffixes
+    # if len(suffixes) < 2 or suffixes[-1] not in (".yml", ".yaml") or source.suffixes[-2] != ".model":
```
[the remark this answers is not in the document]
FynnBe: since the addition of the `type` field to the model spec, the ".model" suffix lost it's importance anyway. Let's drop that check and only leave in checking for a valid yaml suffix....

bioimageio/spec/latest/build_spec.py
```diff
@@ -412,3 +413,31 @@ def build_spec(
...
+    new_weights = _get_weights(weight_uri, weight_type, None, root, **weight_kwargs)[0]
+    model.weights.update(new_weights)
+
+    serialized = spec.schema.Model().dump(model)
+    model = spec.schema.Model().load(serialized)
+
+    return model
+
+
+def serialize_spec(model, out_path, clear_defaults=True):
```
FynnBe, line 436: now, instead of potentially adding default values we potentially remove explicitly set default values... 

I would prefer explicit over implicit and in doubt write out the defaults instead of removing anything...
Maybe we can find a better way forward in https://github.com/bioimage-io/spec-bioimage-io/issues/118 first

FynnBe (changes requested):

bioimageio/spec/v0_3/utils.py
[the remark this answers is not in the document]
constantinpape: So you mean just the following, right?
```python
if suffixes[-1] not in (".yml", ".yaml"):
  raise ValidationError("....")
```

oeway:
Can we only support .yaml not .yml, I read it somewhere said it's recommended to use .yaml.

constantinpape:
> Can we only support .yaml not .yml, I read it somewhere said it's recommended to use .yaml.

Fine with me.

FynnBe:
> Can we only support .yaml not .yml, I read it somewhere said it's recommended to use .yaml. 

fine with me, too, but I would suggest to leave that for 0.4.0.

bioimageio/spec/latest/build_spec.py
```diff
@@ -412,3 +425,19 @@ def build_spec(
...
     return model
+
+
+def add_weights(model, weight_uri: str, root: Optional[str] = None, weight_type: Optional[str] = None, **weight_kwargs):
+    """Add weight entry to bioimage.io model."""
+    new_weights = _get_weights(weight_uri, weight_type, None, root, **weight_kwargs)[0]
+    model.weights.update(new_weights)
+
+    serialized = spec.schema.Model().dump(model)
+    model = spec.schema.Model().load(serialized)
```
FynnBe, line 436: instead of this roundtrip in `add_weights` we could either do the roundtrip for the WeightEntry in `_get_weights`... 
(or here for each weight entry in the update)

FynnBe:
after resolving https://github.com/bioimage-io/spec-bioimage-io/pull/119#discussion_r653614214 this PR should looks ready to merge to me

bioimageio/spec/latest/build_spec.py
constantinpape: Is there a reason not to do the full Roundtrip here? But I am fine with doing the roundtrip just for the weights, can you go ahead and change it?

FynnBe (approved):

PR #122, opened 2021-06-16
Reorganize imports

### LINKED ISSUE #120: import - package structure
after a recent refactor there are 3 potential ways to import the v3 spec (which all come with different limitations, those are _not_ equivalent):

1. ```Python
    import bioimageio.spec.v0_3
    ```
1. ```Python
    import bioimageio.spec.latest
    ```

1. ```Python
    import bioimageio.spec
    ```

I can see that I would do 1. do get a specific spec, and 2. and 3. if I want the latest.

__Proposal 1__: let's get rid of `latest` :) I'm not sure what this should even mean. If you start working on `v0_4`, is this automatically the latest? or does latest mean current stable? Furthermore having it in the "root" of spec should be "latest" enough.

then there are issues how the imports from submodules are passed on. Let's take `bioimageio/spec/__init__.py`:

https://github.com/bioimage-io/spec-bioimage-io/blob/afff3a45a4e214d76b072f61dc6ef4c68a46054e/bioimageio/spec/__init__.py#L1-L4

(latest `__init__` will have `from v0_3 import *`)

with this submodule import is broken:

```Python
# doesn't work :(
from bioimageio.spec import raw_nodes

# works:
import bioimageio.spec
bioimageio.spec.raw_nodes.Something
```
This is because the "linking" via `import *` does not generate modules that can be discovered - it makes attributes available in the module namespace. There simply is no `raw_nodes.py` in the `/spec`, so it isn't discovered.
There are solutions to this:

1. live with it: doing import `bioimageio.raw_nodes` is not supported anymore :)
2. Write a custom finder/loader and hook into the Python import system: Honestly, until yesterday I didn't even know about this. Looks like a cool system to load modules not incarnated in the file system, e.b. a database. [PMOTW](https://pymotw.com/3/sys/imports.html) has a nice example. 
3. autogenerate pseudo modules in `/bioimageio/spec`. These are placeholder modules that `import *` from the relevant module in `v0_3` or whatever the _current_ (not necessarily latest...) spec is. Something like the following does the job:
  ```Python
  # simplified, in fact `__init__.py` has to be modified, too
  in_path = Path("./bioimageio/spec/v0_3")
  out_path = Path("./bioimageio/spec")
  for f in in_path.glob("*.py"):
      if f.name.startswith("__"):
          continue
    tmp_mod = out_path / f.name
    tmp_mod.write_text(f"# Auto-generated by gen.py - do not modify\n\nfrom .v0_3.{f.stem} import *\n")
  ```
From these solutions I have to say I favor 3:
* (1) is just ugly
* (2) is obscure, maintainability is dependent on advanced knowledge of the Python import system. I don't think preferring a non-standard file structure in a file based package justifies it.
* (3) is very easy to understand, it can be automated, checked for in CI. I'd consider maintainability high. It is a custom step, however, probably easy to miss if someone is new to the project.

__Proposal 2__: let's try to fix imports with (3)

CC: @FynnBe @constantinpape 

P.S.: To make matters worse, we need to decide on this quickly - otherwise, I'll have to assume (1) and refactor downstream packages accordingly.

k-dominik:
Summary:

* This PR removes the `bioimageio.spec.latest` module. `gen_spec` is moved to `bioimageio.spec`.
* Added a script to automatically generate "passthrough" modules for a (single) specific version, ran this for `v0_3`. This makes the `v0_3` spec available from `bioimage.spec`.

I kept this as draft, because after actually doing it, I think the approach with the script that generates those modules might be to fragile. Also I feel I would have to document it a lot, which also seems like a sign, that it's a bad idea. Maybe living with the fact that `from bioimageio.spec.raw_nodes import Something` doesn't work is less painful.

fixes #120


constantinpape (approved):
LGTM, but I think it would be good if @FynnBe has another look at this too.

k-dominik:
> LGTM, but I think it would be good if @FynnBe has another look at this too.
> 
> Edit: Sorry, I didn't see that this is still marked as draft, I assumed it was good to go since tests pass.

Thx @constantinpape for the review. I left it a draft because I felt it turned out a bit more complicated than I anticipated and was not sure whether it was a good idea or not. But I think we'd have similar issues with the custom finder/loader approach.

Let's let @FynnBe have a look and then maybe decide.

FynnBe (approved):
LGTM

it would be great to add a CI check and maybe also a pre-commit hook for convenience.

bioimageio/spec/__init__.py
```diff
@@ -1,4 +1,23 @@
...
+download_uri_to_local_path = utils.download_uri_to_local_path
+load_raw_model = utils.load_raw_model
+load_model = utils.load_model
+
+# autogen: stop
+
+# assuming schema will always be part of spec
+from . import schema
+from .raw_nodes import FormatVersion
+__version__ = schema.get_args(FormatVersion)[-1]
```
FynnBe, line 23: here you are actually assuming that schema always imports `typing.get_args`

Did you avoid 
```python
try:
    from typing import get_args
except ImportError:
    from typing_extensions import get_args

```
her ein __init__.py on purpose?

k-dominik: good catch. this doesn't make any sense - I'm going to move that import to `spec.shared.common`, and change all imports accordingly.

Issue #108, opened 2021-06-11
What happened to spec.load_and_resolve_spec?


constantinpape:
@FynnBe we used to have this function to load the spec from a yaml and build the spec object. But now:
```
$ python -c "import bioimageio.spec as spec; spec.load_and_resolve_spec"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'bioimageio.spec' has no attribute 'load_and_resolve_spec'
```
Did you rename this function?

constantinpape:
I guess it's 
https://github.com/bioimage-io/spec-bioimage-io/blob/master/bioimageio/spec/v0_3/utils.py#L55

But when I try it I get the following error (also it's clear from the returnval / function signature of `resolve_raw_node_to_node`):
```
  File "/g/kreshuk/pape/Work/bioimageio/spec-bioimage-io/bioimageio/spec/v0_3/utils.py", line 56, in load_model
    return resolve_raw_node_to_node(*load_raw_model(source, root_path=root_path))
TypeError: resolve_raw_node_to_node() missing 1 required positional argument: 'nodes_module'
```

FynnBe:
yes I renamed it `load_model` (to keep the analogy to raw_nodes/nodes there is load_raw_model/load_model). 
hopefully da2510769c7c36d8034e046fcaa70c571573fd51 solves this. apparently we are lacking some tests that would cover `load_module`...

constantinpape:
Ok, I made it work with the latest commit, but had to fix two more issues:

It looks like "root_path" is not really optional, if I don't pass it I get the following error:
```
Traceback (most recent call last):                                                                                                          
  File "export_bioimageio_model.py", line 104, in <module>                                                                                  
    export_to_bioimageio(args.checkpoint, args.input, args.output,                                                                          
  File "export_bioimageio_model.py", line 84, in export_to_bioimageio                                                                       
    export_biomageio_model(                                                                                                                 
  File "/home/pape/Work/software/conda/miniconda3/envs/torch17/lib/python3.8/site-packages/torch_em/util/modelzoo.py", line 557, in export_b
iomageio_model                                                                                                                              
    val_success = _validate_model(out_path)                                                                                                 
  File "/home/pape/Work/software/conda/miniconda3/envs/torch17/lib/python3.8/site-packages/torch_em/util/modelzoo.py", line 469, in _validat
e_model                                                                                                                                     
[...]

PR #124, opened 2021-06-17
Improve load model

FynnBe:
builds on top of #119 and #122


FynnBe:
fixes #108

constantinpape (commented):
Looks good, but can we add a test for `load_model` before merging?
Judging from past experiences we should have a test for this before merging any relevant changes.

FynnBe:
https://github.com/bioimage-io/spec-bioimage-io/pull/124/commits/d80fba8979d2b1f476769102faa8590a320b9d98 depends on  bioimageio.torch which depends on spec. so adding it as a dependency here is an ugly cycle.

to avoid this in the meantime the test is now conducted with the RF example, which is scheduled to be removed soon, as we do not want to support pickle weights and scikit-learn as a framework...

constantinpape:
> to avoid this in the meantime the test is now conducted with the RF example, which is scheduled to be removed soon, as we do not want to support pickle weights and scikit-learn as a framework...

I think that's fine for now. I would vote for removing the sklearn model and pickle weights in a separate PR where we also move the unet example model here (and can then use it to replace all the rf model in the tests).

constantinpape:
The tests now fail with `   E   ModuleNotFoundError: No module named 'bioimageio.sklearn'`, because we have not moved this part of the repo over from `python-bioimage-io`. Could we instead just call `sklearn.ensemble.RandomForestClassifier` directly to fix this?

FynnBe:
> The tests now fail with ` E ModuleNotFoundError: No module named 'bioimageio.sklearn'`, because we have not moved this part of the repo over from `python-bioimage-io`. Could we instead just call `sklearn.ensemble.RandomForestClassifier` directly to fix this?

yeah... almost the same problem... I basically did that, but we have to wrap the `sklearn.ensemble.RandomForestClassifier` so I added the wrapper as local source in rf.py

setup.py
```diff
@@ -29,6 +29,7 @@
         "python-stdnum",
         "requests",
         "ruamel.yaml",
+        "scikit-learn",
```
k-dominik, line 32: since it was never part of the deps so far, and support for this is also fading, `scikit-learn` could for now be added to the test dependencies only.

k-dominik (commented):
LGTM

PR #129, opened 2021-06-19
Remove pickle weight forma and sklearn example model

constantinpape:
TODO

- [ ] Add unet example model for all format versions
- [ ] Make tests pass


constantinpape:
Superseded by other PRs.

Issue #131, opened 2021-06-21
source spdx_licenses from somwehere else


k-dominik:
Currently, the pypi package `spdx-license-list` is used as a source for valid license identifiers. [That package](https://gitlab.com/uniqx/spdx-license-list) is deprecated (also we'd need to package it's license :o when making our binaries available). As far as I can see we are using the license _keys_ only. I guess we could create a python file containing either just a list, or the full dictionary from [here](https://github.com/spdx/license-list-data/blob/master/json/licenses.json), slapping the license (according to [this](https://github.com/spdx/license-list-data/blob/master/accessingLicenses.md) this should be  CC-BY-3.0) on it and lose the dependency?

FynnBe:
will be solved in #132

PR #132, opened 2021-06-21
replace deprecated spdx-license-list package by adding license specs form spdx source as  json

k-dominik:
The spdx-license-list package is deprecated (#131).

This PR adds the license data from  spdx license list  https://github.com/spdx/license-list-data/releases/tag/v3.13

I think this is easier than creating an extra package and maintaining it.

### DIFF
--- a/bioimageio/spec/shared/__init__.py
+++ b/bioimageio/spec/shared/__init__.py
@@ -1,1 +1,12 @@
+import json
+
+from pathlib import Path
+
 from .common import BIOIMAGEIO_CACHE_PATH, yaml
+
+
+_license_file = Path(__file__).parent.parent / "static" / "licenses.json"
+_license_data = json.loads(_license_file.read_text())
+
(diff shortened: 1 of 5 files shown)


k-dominik:
geez - 5000 lines of dict :/ maybe this mandates putting it somewhere separate after all - what do you think @FynnBe?

FynnBe:
as discussed offline: we'll commit the [raw json](https://spdx.org/licenses/licenses.json) instead. 
(sure it's a lot of lines, but only text after all 😉)

FynnBe:
also, just to keep a record: we decided to only allow for licenceIDs to be specified (for simplicity and to avoid typos, also @k-dominik : conda-forge does the same)

k-dominik:
After a discussion with @FynnBe and @constantinpape we decided to include the `.json` file from the official source, and load it dynamically instead of ingesting it as a python file via a script.

constantinpape (approved):

FynnBe (approved):

PR #133, opened 2021-06-22
Add license file

k-dominik:
are there any opinions on licenses, everyone @bioimage-io/admin ?
I think it's best to keep it very open - after all we want people to use it :)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 BioImage.IO team
+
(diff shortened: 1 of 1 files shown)


oeway:
MIT is a good choice! Thanks for the PR.

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 BioImage.io team
```
oeway, line 3: Maybe make io capital, to be consistent.

constantinpape (approved):
+1 for MIT

Issue #127, opened 2021-06-18
make "passthrough" module check more correct :)


k-dominik:
the script (`scripts/generate_passthrough_models.py`) currently:
1. removes previously autogenerated modules from `bioimageio.spec`,
2. adds new modules in `bioimageio.spec` with links to the _current_ version, and
3. replaces _some_ contents in `bioimageio/spec/__init__.py`.

the `check` subcommand currently only checks for correctness of (2).

The check should also make sure `__init__.py` is enriched with the right content.

FynnBe:
as a pre-commit hook this check could be performed by simply running the generation (it would then "fail" due to changed files). Would that solve this and is there a CI analog?

FynnBe:
a pre-commit hook for this in in place and generate_version_submodule_imports has been improved

PR #139, opened 2021-06-23
add pre-commit hook for passthrough model script

FynnBe:
currently the "generate_passthrough_modules.py" script fails for me with:
```
 line 142, in main
    return args.func(args)
AttributeError: 'Namespace' object has no attribute 'func'
```
@k-dominik any idea what's going wrong here?

I suppose at some point it stopped working.. would be good to add this pre-commit hook to avoid this... 
I'll look into a ci test as well...


k-dominik:
you need to pass some stuff to the script:

```bash
$: python scripts/generate_passthrough_modules.py --help
usage: generate_passthrough_modules.py [-h] {check,generate} ...

script that generates Python module files in bioimageio.spec that 'link' to a certain spec version. The generated modules act as pass-
through, via `from .vX_Y import *

optional arguments:
  -h, --help        show this help message and exit

subcommands:
  {check,generate}
```

so a valid call would be

```bash
$: python scripts/generate_passthrough_modules.py check --target-version v0_3
Checking `bioimageio.spec` modules to link against v0_3.
Assuming module location /home/kutra/sources/tiktorch/vendor/spec-bioimage-io/bioimageio/spec, with target spec in /home/kutra/sources/tiktorch/vendor/spec-bioimage-io/bioimageio/spec/v0_3.
All seems fine.
```

I guess the problem now is that `--target-version` needs to be specified :( (but this also enables you to work on the next version)...

FynnBe:
how about we add the default to check against the 'latest' (aka hightest version).. matching by looking at the folders `v{MAJ}_{MIN}`?

FynnBe:
relates to https://github.com/bioimage-io/spec-bioimage-io/issues/127

k-dominik:
that sounds perfectly sensible! I won't have time to do this before Friday - so if it's okay to leave this one hanging a bit...

FynnBe:
we can leave https://github.com/bioimage-io/spec-bioimage-io/issues/127 open to discuss further improvements, but merge this PR already I think, what do you reconn @k-dominik ?

PR #145, opened 2021-06-29
Mypy

FynnBe:
builds on top of #144 

improves type annotations and add Mypy CI


bioimageio/spec/build_spec.py
```diff
@@ -74,24 +72,19 @@ def _get_weights(weight_uri, weight_type, source, root, **kwargs):
     if weight_type == "pytorch_state_dict":
```
oeway, line 72: I am thinking can you read the weight format definition json file rather than hard code all these info? If necessary we can add more keys in to the weight format definition. What do you think?

PR #146, opened 2021-06-30
Packager

FynnBe:
builds on top of #145 

implements `export_package`


oeway:
I haven't read your implementation but let me add some comments here and hopefully will give you some extra context for the implementation. I think it should have few options to use the python packager: 1) used to produce a zip package directly 2) used to generate a list of files that will be uploaded to zenodo one by one. 3) given an id or a yaml file from bioimage.io (can also via some URI scheme), the python library get this id can pull files locally to run the model 4) similar to 30 but instead of downloading the files, it only probe whether the files exists (by sending HEAD request) in the remote location, this is for CI testing of models.

A nice implementation would be first probe all the files and throw error if necessary, then start the full download. This is needed because some files are large, and it often happens that the user wait for some minutes for the large files finished, the it crash on small missing files.

bioimageio/spec/shared/io.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        if isinstance(source, str) and source.startswith("http"):
+            raise ValueError("Cannot export package for remote source")
+
+        raw_model, root_path = cls.ensure_raw_model(source, root_path, update_to_current_format)
+        io_cls = cls.get_matching_io_class(raw_model.format_version, "export")
+        package_path = io_cls._make_package_wo_conversions(
+            raw_model, root_path, weights_formats_priorities=weights_formats_priorities
+        )
+        return package_path
```
FynnBe, line 140: > 1) used to produce a zip package directly

this class method is exposed as `export_package` for every subversion (0.1, 0.3) and returns a path to the zipped package in the systems <tmp>/bioimageio_cache folder

bioimageio/spec/v0_3/__init__.py
```diff
@@ -1,12 +1,15 @@
...
 
-get_nn_instance = utils.get_nn_instance
-download_uri_to_local_path = utils.download_uri_to_local_path
-load_raw_model = utils.load_raw_model
-load_model = utils.load_model
+load_raw_model = utils.IO.load_raw_model
+load_model = utils.IO.load_model
+save_raw_model = utils.IO.save_raw_model
+serialize_raw_model_to_dict = utils.IO.serialize_raw_model_to_dict
+export_package = utils.IO.export_package
```
FynnBe, line 13: current version of `export_package` exposed at `bioimageio.spec.export_package` 
thanks to @k-dominik this is autogenerated from `bioimageio.spec.v0_3.export_package`

bioimageio/spec/shared/io.py
```diff
@@ -0,0 +1,348 @@
...
+
+    @classmethod
+    def get_package_content(
+        cls,
+        source: Union[RawModelNode, os.PathLike, str, dict],
+        root_path: pathlib.Path,
+        update_to_current_format: bool = False,
+        weights_formats_priorities: Optional[Sequence[str]] = None,
+    ) -> Dict[str, Union[str, pathlib.Path]]:
```
FynnBe, line 172: > 2) used to generate a list of files that will be uploaded to zenodo one by on

`get_package_content()` returns a dict of file names and their current local path (or their content as a string)

FynnBe:
@oeway thanks for your input!
I'll add 3) and 4) as well...

example_specs/models/unet2d_nuclei_broad/rdf_v0_1_0.yaml
```diff
@@ -0,0 +1,99 @@
...
+test_input: test_input.npy
+test_output: test_output.npy
+
+prediction:
+  dependencies: conda:./environment.yaml
+  weights:
+     source: https://zenodo.org/record/3446812/files/unet2d_weights.torch
+     hash: {sha256: e4d3885bccbe41cbf6c1d825f3cd2b707c7021ead5593156007e407a16b27cf2}
+
+config:
```
k-dominik, line 54: I think it would be worth leaving a comment here, what this future stuff is about. `convert_model_from_v0_1` has no docstring, I guess these values are intended to be used there?

bioimageio/spec/v0_3/raw_nodes.py
```diff
@@ -2,11 +2,12 @@
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, NewType, Tuple, Union
+from typing import Any, ClassVar, Dict, List, NewType, Tuple, Union
 
 from marshmallow import missing
 from marshmallow.utils import _Missing
 
+from bioimageio.spec import v0_1
```
k-dominik, line 10: should this then be maybe moved to common? I wouldn't expect v0_1 imports here.

k-dominik (commented):
I have to admit that I didn't have time to look into all changes to a level that I could confidently say I understand why you implemented certain things in a certain way.

Changes improve readability to the incarnation before. Method/class/function/variable names are quite expressive. Still, I'd miss some comments with motivations. Comments are in general sparse (and often in places where it is at least for me it is totally clear whats happening, but maybe I'm biased to remember those.).

I think in future there should be some sort of developer documentation :), too, that explains the broad concepts and motivations in this library a bit.

Minor comment: This is a cumulative PR and I understand that it is hard to summarize this. I went through all parents and found that also there, summaries were also quite short. Something like "Changes motivated by _this_long_issue" do not really help a lot. Nice to have those links for details, but I see having a meaningful summary to be a minimum requirement for a PR.

bioimageio/spec/shared/io.py
```diff
@@ -0,0 +1,348 @@
...
+    def _get_package_content_wo_conversions(
+        cls,
+        raw_model: current_spec.raw_nodes.Model,
+        root_path: pathlib.Path,
+        weights_formats_priorities: Optional[Sequence[str]],
+    ) -> Dict[str, Union[str, pathlib.Path]]:
+        package = NoOverridesDict(
+            key_exists_error_msg="Package content conflict for {key}"
+        )  # todo: add check in model validation
+        package["original_rdf.txt"] = cls.serialize_raw_model(raw_model)
```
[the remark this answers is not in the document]
FynnBe: yes, but we have to solve the possible existence of a second .yaml file in the model package in general (could be a a file attachment, etc..., but requiring the file name `rdf.yaml` in a package (outside of a package any name can be allowed)
```suggestion
        package["original_rdf.txt"] = cls.serialize_raw_model(raw_model)  # todo: .txt -> .yaml once 'rdf.yaml' is only valid rdf file name in package
```

bioimageio/spec/v0_1/raw_nodes.py
```diff
@@ -1,3 +1,103 @@
-from bioimageio.spec.v0_3.raw_nodes import *
+"""
+raw_nodes 0.1 only implements the Model class (and its requirements) and not other aspects of the deprecated 0.1 spec
```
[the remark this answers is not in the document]
FynnBe: for 0.1 there is definitely a blurred line between general and model RDF. at that time we also defined specifications for Reader, Sampler, etc. which have been dropped completely for now. This aspect is not in use anywhere, (any 0.1 RDFs really), so I left it out and only included the 0.1 model RDF for completeness and to develop the IO and model conversion mechanics.
```suggestion
raw_nodes 0.1 only implements the Model class (and its requirements) and not other aspects of the deprecated 0.1 RDFs like 'Reader', 'Sampler' and 'Transformation'
```

tests/test_dump_spec.py
```diff
@@ -1,24 +1,25 @@
...
+    raw_model = load_raw_model(unet2d_nuclei_broad_any_minor_path)
 
-    serialized = schema.Model().dump(raw_model)
+    serialized = serialize_raw_model_to_dict(raw_model)
     assert isinstance(serialized, dict)
 
-    # yaml.dump(serialized, Path() / "serialized.yml")
+    # from pathlib import Path
+    # from bioimageio.spec import save_raw_model
+    # save_raw_model(raw_model, Path() / "serialized.yml")
```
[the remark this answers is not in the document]
FynnBe: somewhat yes, but left in on purpose to show how to serialize to file (and for future debugging)...

bioimageio/spec/v0_3/raw_nodes.py
FynnBe: why not? v0_3 can reuse definitons from v0_1, no?

As there is a possiblity to convert from 0.1.x to 0.3.x format version is defined below as:
```
ModelFormatVersion = Literal[
    v0_1.ModelFormatVersion, "0.3.0", "0.3.1", "0.3.2"  # newest format needs to be last (used in spec.__init__.py)
]
```

I changed this as the current 'root' spec is the latest, which should include previous versions to the extent that is supported. E.g. we might break this chain with major release bumps.

example_specs/models/unet2d_nuclei_broad/rdf_v0_1_0.yaml
FynnBe: yes, I will add documentation for this

bioimageio/spec/v0_1/converters.py
```diff
@@ -0,0 +1,6 @@
+def maybe_convert_model(data):
```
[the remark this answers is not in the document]
FynnBe:
```suggestion
def maybe_convert_model(data):
    """the only v0_1 format version is 0.1.0. As there is no preceding format version, no conversion is applied"""
```

bioimageio/spec/shared/io.py
```diff
@@ -0,0 +1,348 @@
...
+            )
+        elif data_version_wo_patch == current_version_wo_patch:
+            return cls
+        elif cls.preceding_io_class is None:
+            raise NotImplementedError(f"format version {'.'.join(map(str, data_version_wo_patch))}")
+        else:
+            return cls.preceding_io_class
+
+    @classmethod
+    def make_zip(cls, path: pathlib.Path, content: Dict[str, Union[str, pathlib.Path]]):
```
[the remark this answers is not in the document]
FynnBe: added in https://github.com/bioimage-io/spec-bioimage-io/pull/181

PR #149, opened 2021-07-05
use 'general RDF' everywhere (not 'generic RDF)

FynnBe:
in the general/generic RDF format documentations we use 'general RDF': https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/rdf_spec_latest.md

I have no strong preference of 'general RDF' over 'generic RDF'. Either version has their context in which they read better, but I think it would be best to stick to one formulation to avoid confusion.

@oeway @k-dominik @constantinpape @esgomezm time for some strong opinions 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Specifications for BioImage.IO
 
-This repository contains specifications defined by the BioImage.IO community. These specifications are used for defining fields in YAML files which we called `Resource Description Files` or `RDF`. The RDFs can be downloaded or uploaded to the [bioimage.io website](https://bioimage.io), produced or consumed by BioImage.IO-compatible consumers(e.g. image analysis software or other website). Currently we defined two types of RDFs: a dedicated RDF specification for AI models (i.e. `model RDF`) and a generic RDF specification. The model RDF is a RDF with additional fields that specifically designed for describing AI models.
+This repository contains specifications defined by the BioImage.IO community. These specifications are used for defining fields in YAML files which we called `Resource Description Files` or `RDF`. The RDFs can be downloaded or uploaded to the [bioimage.io website](https://bioimage.io), produced or consumed by BioImage.IO-compatible consumers(e.g. image analysis software or other website). Currently we defined two types of RDFs: a dedicated RDF specification for AI models (i.e. `model RDF`) and a general RDF specification. The model RDF is a RDF with additional fields that specifically designed for describing AI models.
 
 
 All the BioImage.IO-compatible RDF must fulfill the following rules:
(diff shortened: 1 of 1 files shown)


constantinpape:
I think general fits better: 
> general is including or involving every part or member of a given or implied entity, whole etc; as opposed to (specific) or (particular) while generic is very comprehensive; pertaining or appropriate to large classes or groups as opposed to specific

(from https://wikidiff.com/general/generic.)

k-dominik:
naming things! Fun! :) Not very strong opinion here, btw.

_Generic_ and _General_ - when I first read the docs here, it wouldn't have made much of a difference to be honest. But that all under the pretext that I am not an RDF nerd. RDF hierarchies should be common - is there a lingo related there? I couldn't really find anything. Otherwise I'd vote for something like `BioimageIO base RDF` :o or anything more descriptive, that makes it clear that this is _not_ a generic, nor _general_ RDF at all. Or to phrase it differently: look at `BioimageIO General RDF`, `BioimageIO Generic RDF` - to me both don't fly as they seem be contradictory (the first part - bioimageIO -> specific, the second part tries to make it really general).

again, not very strong opinion.

FynnBe:
thanks @constantinpape , @k-dominik interesting feedback!

> Otherwise I'd vote for something like `BioimageIO base RDF` :o or anything more descriptive, that makes it clear that this is _not_ a generic, nor _general_ RDF at all

to me 'base RDF' would imply that it is what bioimageio RDFs have in common, but that it is not usable on its own, which it is though, as it covers (atm) datasets, notebooks, collections and applications.


> (the first part - bioimageIO -> specific, the second part tries to make it really general).

well, it is bioimageio specific... but covers several use cases (see above)

esgomezm:
I agree with @constantinpape that general sounds better

FynnBe:
> I agree with @constantinpape that general sounds better

awesome! I'll merge this then (won't be on master right way, as it's now part of #146 ...)

oeway:
Haha, fun thing again! I think there is no big difference. however, I do see in Python, we define types as `generic` types instead of general types: https://docs.python.org/3/library/typing.html


Well, when I read the link posted by @constantinpape  I feel the opposite.

> general is including or involving every part or member of a given or implied entity

Well, the generic/general RDF doesn't include every part, e.g. `inputs/outputs` keys are not included. But of cause the parts in the RDF itself is included in specific RDFs.


When you read also the definition of `generic`, it seems also fit:

> Lacking in precision, often in an evasive fashion; vague; imprecise.
(of a product or drug) Not having a brand name.
(biology, not comparable) Of or relating to a taxonomic genus.
(grammar) Specifying neither masculine nor feminine; epicene.

FynnBe:
> Haha, fun thing again! I think there is no big difference. however, I do see in Python, we define types as `generic` types instead of general types: https://docs.python.org/3/library/typing.html

That's a neat example for the use of 'generic', but a different scenario as typing.Generic is all about being parametrized with another type, e.g. typing.List is a generic datatype that allows `List[str]`, `List[int]`, etc... 
With this in mind I think 'general RDF' fits even better as it is not generic in this sense.

oeway:
@FynnBe It seems you forgot to replace the `generic RDF` in the README file.

FynnBe:
> @FynnBe It seems you forgot to replace the `generic RDF` in the README file.

I might be missing something, but this is all that this PR does (replace 'generic RDF' in README.md: https://github.com/bioimage-io/spec-bioimage-io/pull/149/commits/b4df7b471fad74cb9c596fece1d6dc11a5c4a06b

oeway:
Strange, did I messed up with the README, it still has generic RDF, see here: https://github.com/bioimage-io/spec-bioimage-io#resource-description-file-specification

FynnBe:
> Strange, did I messed up with the README, it still has generic RDF, see here: https://github.com/bioimage-io/spec-bioimage-io#resource-description-file-specification

yes, this change has not propagated to main yet and is now part of #146

PR #150, opened 2021-07-05
Packager cli

FynnBe:
1. use IO class for all (=for now collection and model) nodes. This works nicely by relying more on the 'type' field (which we then probably never ever should get rid of again^^) (@constantinpape the IO_Base part might be a bit confusing as @k-dominik already noted in #146 . I added a little bit of documentation, but still not enough. Feel free to leave this part for @k-dominik to review)
2. for the [CLI update](https://github.com/bioimage-io/spec-bioimage-io/blob/packager_cli/bioimageio/spec/__main__.py) I would appreciate both of your feedback.


bioimageio/spec/__main__.py
```diff
@@ -1,96 +1,64 @@
...
 
 yaml = YAML(typ="safe")
 
 app = typer.Typer()  # https://typer.tiangolo.com/
 
 
 @app.command()
-def package_model(model_yaml: Path, auto_convert: bool = False):
-    pass
+def package(rdf_source: str, auto_convert: bool = False):
```
constantinpape, line 16: I guess this is being implemented in #147?

bioimageio/spec/__main__.py
```diff
@@ -1,96 +1,64 @@
...
-            pprint(e.messages)
-            code = 1
-        except Exception as e:
-            print("invalid model:", model["source"])
-            pprint(e)
-            code = 1
 
-    return code
+@app.command()
+def verify_spec(model_yaml: str, auto_convert: bool = False):
```
constantinpape, line 53: Can we add some doc strings / help for this? I.e. `verify_spec --help` should print something useful.

bioimageio/spec/__main__.py
```diff
@@ -1,96 +1,64 @@
...
+    warnings.warn("'verify_spec' is deprecated in favor of 'verify'")
+    return verify(model_yaml, auto_convert)
 
 
 @app.command()
 def verify_bioimageio_manifest(manifest_yaml: Path, auto_convert: bool = False):
```
constantinpape, line 59: Same as `verify_spec`.

constantinpape (commented):
I only had a quick look at the new CLI (and more doc there would be good). In general, these PRs are too big to properly review for me rn. I think it would be good to increase test coverage (and maybe also add code cov to make this more explicit) to make sure that large design changes are sound.

FynnBe:
just stumbled upon us always talking about the 'validator', yet having named the command 'verify'...

the difference of validation and verification is a bit confusion...
> verification: “A test of a system to prove that it meets all its specified requirements at a particular stage of its development.” - IEEE-STD-610 
https://www.plutora.com/blog/verification-vs-validation

> validation: “An activity that ensures that an end product stakeholder’s true needs and expectations are met.” -- IEEE-STD-610
https://www.plutora.com/blog/verification-vs-validation

... but I think we should name the command `validate`. Any objections?

bioimageio/spec/__main__.py
FynnBe: the backend is, just missing compression (https://github.com/bioimage-io/spec-bioimage-io/pull/146#discussion_r662778485)

I'll make another PR to #146 for this to keep them small

constantinpape:
I agree, `validate` is more appropriate.

FynnBe:
I think 'update-format' captures the intend of this flag batter than the somewhat ambiguous 'auto-convert'.

bioimageio/spec/__main__.py
FynnBe: opened #152

PR #151, opened 2021-07-05
Add CI for publishing on PyPI

oeway:
This PR add a few github actions workflow to label issues and publish package automatically to pypi. 

Note that you should change the `VERSION` file to bump versions after this PR, if there is any change in the version number, github CI will automatically publish to twitter.

One question though, if we run `bioimageio.spec.__version__` it currently print the spec version, which I think is not expected, because it should print the package version instead. I suggest we use `bioimageio.spec.format_version` for the spec version, and keep `__version__` as the python package version.

Also currently the version number is 0.3b, and when I publish to pypi it will become 0.3b0, so I will change it to 0.3b0 instead.
Is there any reason to not use semantic versioning for the package? 

@FynnBe


oeway:
The test only runs on main branch, so I will merge and see how it goes.

oeway:
I had to commit a few other changes directly to fix the release action, but now it's intact. It will label issues and draft release. Once we change the version number it will publish it to pypi.

The CI actions are taken from @MartinHjelmare

constantinpape:
> Also currently the version number is 0.3b, and when I publish to pypi it will become 0.3b0, so I will change it to 0.3b0 instead.

0.3b sounds wrong, we should use the same version as in the gh releases (currently 0.3.2)

FynnBe:
the `0.3b` is very old and was just meant as a placeholder....

FynnBe:
we should definitely publish as 0.3.2. (once that is finalized). 
About moving version tags...
What would be an appropriate addendum if we were to move a such a version tag again? 0.3.2p1 ?? or do we then just bump patch to 0.3.3, even though in this hypothetical example there would not be any change to the spec from 0.3.2 to 0.3.3?

oeway:
No, I was try to say that we should differentiate the format version of the spec and the version of the python implementation, of course we can make the python package version the same as the format version but adding suffix to it, but I want to confirm this. What do you think @FynnBe @constantinpape .

FynnBe:
would be very straigt forward to have a python package for each released version though... 

from https://semver.org/#semantic-versioning-specification-semver I gather that there is no way of specifying a version larger than 0.3.2 (placeholder for spec/model format version) that begins with 0.3.2 (all pre-release versions make it 'younger' and build versions are ignored ("Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence") 

So maybe the 0.3.2 could be in build altogether instead and the package gets its own version?

Or, we always release as 'pre-release'. I think then we can have `0.3.2-r0` < ` 0.3.2r1` < ...  < 0.3.2 (hypothetical, never to actually be released)

oeway:
> So maybe the 0.3.2 could be in build altogether instead and the package gets its own version?

That's not possible in practice, even if the spec has never changed, we will always have implementation issues need to be fixed and released as a package.

> Or, we always release as 'pre-release'. I think then we can have 0.3.2-r0 < 0.3.2-r1 < ... < 0.3.2 (hypothetical, never to actually be released)
edit: the 'r0', 'r1' parts are chosen arbitrary, any [0-9A-Za-z] combination is allowed

Any of these would work, maybe we can combine `a0`, `b0` and `rc0`(stands for release candidate 0)? see here: https://www.python.org/dev/peps/pep-0440/#pre-releases

FynnBe:
> Any of these would work, maybe we can combine a0, b0 and rc0(stands for release candidate 0)? see here: https://www.python.org/dev/peps/pep-0440/#pre-releases

all fine with me, although if we do stick with this and never actually release the 'final release' of a version 'rc' might be misleading. (that's why I chose r for 'release'). but 'rc' would be just fine for me, too.

constantinpape:
> course we can make the python package version the same as the format version but adding suffix to it, but I want to confirm this. What do you think @FynnBe @constantinpape .

Yes, I think that  is the right way to go!

I don't have a strong opinion about the release suffixes, but I agree with @FynnBe that `rc` could be a bit confusing in this context.

PR #155, opened 2021-07-08
Fix the restriction on type key for general RDF

oeway:
In the general RDF, the type should not be enum but rather any string would be valid.

See the original definition here: https://github.com/bioimage-io/bioimage.io/blob/0002e32e587a9dcd588e3c83981ec4b80a93f835/docs/contribute_models/resource-description-file.md

### DIFF
--- a/bioimageio/spec/v0_3/raw_nodes.py
+++ b/bioimageio/spec/v0_3/raw_nodes.py
@@ -37,7 +37,7 @@ PreprocessingName = Literal["binarize", "clip", "scale_linear", "sigmoid", "zero
 PostprocessingName = Literal[
     "binarize", "clip", "scale_linear", "sigmoid", "zero_mean_unit_variance", "scale_range", "scale_mean_variance"
 ]
-Type = Literal["model", "dataset", "application", "notebook"]
+Type = NewType("Type", str)
 WeightsFormat = Literal[
     "pytorch_state_dict", "pytorch_script", "keras_hdf5", "tensorflow_js", "tensorflow_saved_model_bundle", "onnx"
 ]
--- a/bioimageio/spec/v0_3/raw_nodes.py
+++ b/bioimageio/spec/v0_3/raw_nodes.py
@@ -97,8 +97,6 @@ class RDF(Node):
         if self.type is missing:
             self.type = self.__class__.__name__.lower()  # noqa
 
-        assert self.type in get_args(Type)
-
 
 @dataclass
 class Preprocessing:
--- a/bioimageio/spec/v0_3/raw_nodes.py
+++ b/bioimageio/spec/v0_3/raw_nodes.py
@@ -37,7 +37,7 @@ PreprocessingName = Literal["binarize", "clip", "scale_linear", "sigmoid", "zero
 PostprocessingName = Literal[
     "binarize", "clip", "scale_linear", "sigmoid", "zero_mean_unit_variance", "scale_range", "scale_mean_variance"
 ]
-Type = NewType("Type", str)
+Type = str
 WeightsFormat = Literal[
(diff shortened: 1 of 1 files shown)


FynnBe:
> see the original definiion

"type of the resource, e.g. model, dataset, application or notebook"

This is not stating that any string should be valid; it is rather undedefined what `type` is exactly...

What is your motivation for allowing any string?

Could we instead add other defined type values to keep it concise?

constantinpape:
I am also not sure we should allow anything here; what would the website do if something is `type: foobar`?

oeway:
> What is your motivation for allowing any string? Could we instead add other defined type values to keep it concise?

The first motivation is to accurately implement what we defined in the RDF 0.2.0, where we defined `type` as:
>  type of the resource, e.g. model, dataset, application or notebook

Note that we only listed a few example, it can be anything.

The second motivation is that when we want to experiment new types, e.g. now we want to add `collections` and `workflow` just for experimental purpose, even if we are completely compatible with the RDF spec, the validator will make it fail.

This constrains we adding anything, imagine later every time when we want to add a new type, we will force to bump the version of general RDF, and the version change need to propagate to all the child RDF. It would be really bad if the website want to add a new type and every consumer software will need to upgrade the model RDF version to keep it updated. I really don't like the entanglement of everything.

I see the general RDF spec a bit more beyond bioimage.io, which is why I am using it for other projects too, there we can easily define new types, but it will become irrelevant to bioimage.io, but nevertheless, it would be good to share the validator. 

I would really make the validator a bit more generic, and make it configurable as I proposed here: https://github.com/bioimage-io/spec-bioimage-io/issues/156

So we don't need to bump version when a new type is added.

> I am also not sure we should allow anything here; what would the website do if something is type: foobar?

The website will ignore it, but I didn't hardcode the types in the website itself, it's loaded from a [config file](https://github.com/bioimage-io/bioimage.io/blob/da2e55ca3d2fd28ac2011b24c7631c3c2833939d/site.config.json#L129-L152), so when we have a new type, we edit the config file.

oeway:
@FynnBe Maybe you know, what is this CI error about `AttributeError: type object 'RDF' has no attribute 'authors'`?

bioimageio/spec/v0_3/raw_nodes.py
```diff
@@ -33,7 +33,7 @@
 PostprocessingName = Literal[
     "binarize", "clip", "scale_linear", "sigmoid", "zero_mean_unit_variance", "scale_range", "scale_mean_variance"
 ]
-Type = Literal["model", "dataset", "application", "notebook"]
+Type = NewType("Type", str)
```
FynnBe, line 36:
```suggestion
Type = str
```

As we do not create every type value by calling `Type("type string")`

FynnBe (changes requested):

FynnBe:
> @FynnBe Maybe you know, what is this CI error about `AttributeError: type object 'RDF' has no attribute 'authors'`?

a rebase on the current `main` should fix this

oeway:
> a rebase on the current `main` should fix this

Did a rebase, but got some other error now, @FynnBe .

FynnBe:
I think it might be easier for us to leave the Type annotated as a Literal (keeping in mind the primary use-cases and our future plans) while just not raising other values. #168 implements that.

oeway:
> I think it might be easier for us to leave the Type annotated as a Literal (keeping in mind the primary use-cases and our future plans) while just not raising other values. #168 implements that.

I think it's an implementation details here, we need the strict restrictions on the types, however, I would rather do it in the website, not here. The supported types are determined by the website, not the python package. I have other websites running the same thing and there we have different supported custom types which does not make sense to submit to the bioimage.io. If we really want to make the validator useful, I would suggest we let website/users/consumer software decide what types are supported.

oeway:
@FynnBe Thanks, it passes now, I will merge it now.

PR #157, opened 2021-07-08
Remove pickle

constantinpape:
@FynnBe I think pickle format can be removed; it looks like it's removed everywhere else now:
```
$ grep -Inr pickle bioimageio/spec/
bioimageio/spec/v0_3/schema.py:781:            "pickle": True,  # todo: remove
```


FynnBe:
yes, nice catch of this lost line 👍

Issue #160, opened 2021-07-08
CLI does not work


constantinpape:
Using the CLI command does not work:
```
$ bioimageio
Traceback (most recent call last):
  File "/home/pape/Work/software/conda/miniconda3/envs/torch19/bin/bioimageio", line 33, in <module>
    sys.exit(load_entry_point('bioimageio.spec', 'console_scripts', 'bioimageio')())
  File "/home/pape/Work/software/conda/miniconda3/envs/torch19/bin/bioimageio", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/home/pape/Work/software/conda/miniconda3/envs/torch19/lib/python3.9/importlib/metadata.py", line 79, in load
    return functools.reduce(getattr, attrs, module)
AttributeError: module 'bioimageio.spec' has no attribute 'main'
```

The problem is that the entry point https://github.com/bioimage-io/spec-bioimage-io/blob/main/setup.py#L40 does not exist.

There is a nice blogpost on how to do this properly with typer:
https://jrwalk.github.io/pages/projects/python-cli-utilities-poetry-typer

FynnBe:
nice, I'll add this right away. It does boil down to replacing
```python
if __name__ == "__main__":
    app()
```
    
 with

```python
def main():
    app()
```

PR #162, opened 2021-07-08
fix bioimageio entry point

FynnBe:
implmentation following https://github.com/bioimage-io/spec-bioimage-io/issues/160

There is no limit on ways how to fix this actually... 

I think it is neat to leave all typer tuff in __main__.py, thus the package itself does not even depend on typer. one can `import bioimageio.spec` without typer :-).. 
having the entry point defined in __main__.py means that our entry_points has to point there (instead of the plain module "__init__.py"). the name of the entry point function does not need to be 'main', so I kept it as `app`, which works just as well.

If any other implementation is preferred by anyone, please feel free to change it. I don't care how exactly this is solved, it should just work 💯

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
         "typer",
         "typing-extensions",
     ],
-    entry_points={"console_scripts": ["bioimageio = bioimageio.spec:main"]},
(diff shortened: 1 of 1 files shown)


constantinpape (approved):
Works now. Awesome!

PR #164, opened 2021-07-09
make sure to include static data in source dist

k-dominik:
Disclaimer: I couldn't reproduce the original issue locally, but our [current (0.3.2.rc0)
dist on pypi](https://files.pythonhosted.org/packages/29/e6/52646eb12b036ac83fd3d3ea0eb8a759d1e947be54755a71c1dc86cb07f8/bioimageio.spec-0.3.2rc0.tar.gz) does not include static/licenses.json

Reading about it, it looks like

* `include_package_data` will will include data from MANIFEST.in, which we don't have. SO at least that I removed.

Also note that the wheel seems to be intact.

@oeway do you have a good idea on how to debug this on the ci?

<details>

```bash
$: tar -xvf ~/Downloads/bioimageio.spec-0.3.2rc0.tar.gz 
bioimageio.spec-0.3.2rc0/
bioimageio.spec-0.3.2rc0/LICENSE
bioimageio.spec-0.3.2rc0/MANIFEST.in
bioimageio.spec-0.3.2rc0/PKG-INFO
bioimageio.spec-0.3.2rc0/README.md
bioimageio.spec-0.3.2rc0/bioimageio/
bioimageio.spec-0.3.2rc0/bioimageio/spec/
bioimageio.spec-0.3.2rc0/bioimageio/spec/VERSION
bioimageio.spec-0.3.2rc0/bioimageio/spec/__init__.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/__main__.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/build_spec.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/commands.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/converters.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/exceptions.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/nodes.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/raw_nodes.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/schema.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/shared/
bioimageio.spec-0.3.2rc0/bioimageio/spec/shared/__init__.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/shared/common.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/shared/field_validators.py
bioimageio.spec-0.3.2rc0/bioimageio/spec/shared/fields.py
[...]

PR #165, opened 2021-07-09
add static data via manifest

k-dominik:
in the _source_ distribution of `bioimageio.spec` 0.3.2.rc0, the `static/licenses.json` file was not included, as there seem to be issues with having both `MANIFEST.in` and  in setup `package_data`.

Add all data the same way.

Also this bumps the version to `0.3.2.rc1` in order to check that it's working :)

superseeds #164

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include bioimageio/spec/VERSION
 include README.md
 include LICENSE
+include bioimageio/spec/static/licenses.json
--- a/bioimageio/spec/VERSION
+++ b/bioimageio/spec/VERSION
@@ -1,3 +1,3 @@
 {
-    "version": "0.3.2.rc0"
+    "version": "0.3.2.rc1"
 }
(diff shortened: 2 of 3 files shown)


k-dominik:
hey @oeway,

could you have a look? That way we will be able to build the conda package from the pip one :)

oeway:
> hey @oeway,
> 
> could you have a look? That way we will be able to build the conda package from the pip one :)

Sure, I am on it now.

oeway:
@k-dominik Looks good to me, I installed it locally and I can find the license.json file in site-package. Do we have a test case for checking the license, if not I would recommend to have one to make sure we don't break it next time.

oeway (approved):

k-dominik:
Thank you for having a look @oeway!

> Do we have a test case for checking the license

we do have https://github.com/bioimage-io/spec-bioimage-io/blob/main/tests/test_licenses.py // this would fail if that file was not present. However that would not fix the issue of it not being the _source_ distribution... But at least we'll notice on conda recipe rebuild (once it's on conda-forge). I consider this a quirk we had before (most likely due to having both `include_package_data` and `package_data` present in conjunction with _some_ version of setuptools...

oeway:
OK, sounds good then! Let's merge it!

PR #161, opened 2021-07-08
version bump 0.3.2.rc0

k-dominik:
I guess we could use this PR to try to discuss/establish a release workflow. We probably don't need to make a big discussion out of this.

There are always options, of course:

1) Change `VERSION` file, Open PR, wait for PR to be approved by (someone specific, N reviewers?), merge, tag - build on CI
2) Discuss (somewhere, BioimageIO meetings?), change `VERSION` file, Commit directly to master - tag - build on CI
3) get rid of version file, use something like versioneer to update version directly from tag. workflow: tag - build on CI.

by opening this PR you can probably see that I'm slightly in favor of 1)

Once this is merged, I'll remove the bare `0.3.2` tag (and rename `0.3.1` and `0.3.0` tags to `x.y.z-release0`)

closes #114

PR #168, opened 2021-07-12
Lift restriction on RDF.type values

FynnBe:
we define 0.3.2 less strict and plan to get stricter with 0.4.0

supersedes #155

### DIFF
--- a/bioimageio/spec/v0_3/schema.py
+++ b/bioimageio/spec/v0_3/schema.py
@@ -196,11 +196,12 @@ E.g. the citation for the model architecture and/or the training data used."""
 
     type = fields.String(required=True)
 
-    @validates("type")
-    def validate_type(self, value):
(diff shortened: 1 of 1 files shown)


oeway:
Thanks! One question though, we have the 0.2.0 which is a direct translation of the previous spec written in markdown, then now we 0.3.0 because we changed the `authors`, right? Will you have both version implemented? Can we decouple the code and potentially make the general RDF spec an independent submodule?

FynnBe:
> Thanks! One question though, we have the 0.2.0 which is a direct translation of the previous spec written in markdown, then now we 0.3.0 because we changed the `authors`, right? Will you have both version implemented? Can we decouple the code and potentially make the general RDF spec an independent submodule?

yes, I outlined this in https://github.com/bioimage-io/spec-bioimage-io/pull/161#issuecomment-876567568 and think it could help clarify their differences a bit...

PR #169, opened 2021-07-12
update readme with conda install

k-dominik:
Added a badge to conda, updated the installation instructions with `conda`, and also updated to new `validate` command instead of `verify-spec`.


oeway (approved):
Looks good! Thanks!

PR #173, opened 2021-07-13
Improve cli

FynnBe:
set __version__ attribute to package version (was model format version before) and print this package version when using the CLI

### DIFF
--- a/bioimageio/spec/__init__.py
+++ b/bioimageio/spec/__init__.py
@@ -1,9 +1,8 @@
+import json
 import pathlib
 
 from . import v0_1, v0_3
 from .build_spec import build_spec
-
-from bioimageio.spec.shared import yaml
 
 # autogen: start
 from bioimageio.spec.shared import fields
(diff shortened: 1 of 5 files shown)


bioimageio/spec/__init__.py
```diff
@@ -16,6 +20,6 @@
 save_raw_node = utils.IO.save_raw_node
 serialize_raw_node_to_dict = utils.IO.serialize_raw_node_to_dict
 
-__version__ = _get_args(ModelFormatVersion)[-1]
-
 # autogen: stop
+
+__version__ = yaml.load(pathlib.Path(__file__).parent / "VERSION")["version"]
```
oeway, line 25: It's a JSON file, so we should use `json.load`, or would `yaml.load` also work?

FynnBe: as it happens to also be valid yaml syntax `yaml.load` also works... But if it fundamentally is a json I'll better change it already...

oeway: Interesting, so yaml is a super set of json? didn't know you can load it with yaml.

FynnBe: almost. there are some differences.. https://stackoverflow.com/a/21585224

oeway (approved):
Looks good then, thanks!

PR #186, opened 2021-07-15
Update generate_json_specs.py

FynnBe:

PR #187, opened 2021-07-15
Fix json shcema filename

oeway:


oeway:
Let's use https://github.com/bioimage-io/spec-bioimage-io/pull/186 instead.

PR #190, opened 2021-07-19
Simplify source import

constantinpape:
This PR simplifies the source import script.

@FynnBe I don't really understand the motivation for the old code and as far as I can see the complicated `f"user_imports{uuid}"` construction is not necessary. However, when using this it lead to some issues with the torchscript export. I don't have a good explanation for why this happens, maybe some nameclashes with `user_imports`?! In any case, the simplified code here fixes the issue.

### DIFF
--- a/bioimageio/spec/shared/utils.py
+++ b/bioimageio/spec/shared/utils.py
@@ -182,7 +182,7 @@ class SourceNodeTransformer(NodeTransformer):
     @staticmethod
     def transform_ResolvedImportableSourceFile(node: ResolvedImportableSourceFile) -> nodes.ImportedSource:
         module_path = resolve_uri(node.source_file)
-        module_name = os.path.splitext(os.path.split(module_path)[1])[0]
+        module_name = f"module_from_source.{module_path.stem}"
         importlib_spec = importlib.util.spec_from_file_location(
             module_name, module_path
         )
--- a/bioimageio/spec/shared/utils.py
+++ b/bioimageio/spec/shared/utils.py
@@ -5,7 +5,6 @@ import os
 import pathlib
 import sys
 import typing
-import uuid
 import warnings
 from functools import singledispatch
 from types import ModuleType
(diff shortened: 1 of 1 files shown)


bioimageio/spec/shared/utils.py
```diff
@@ -182,8 +181,10 @@ def transform_ImportableModule(node):
 
     @staticmethod
     def transform_ResolvedImportableSourceFile(node: ResolvedImportableSourceFile) -> nodes.ImportedSource:
+        module_path = resolve_uri(node.source_file)
+        module_name = os.path.splitext(os.path.split(module_path)[1])[0]
         importlib_spec = importlib.util.spec_from_file_location(
-            f"user_imports.{uuid.uuid4().hex}", resolve_uri(node.source_file)
+            module_name, module_path
```
FynnBe, line 187: this might collide with an existing module...
How is the random module name of  `uuid.uuid4().hex` causing any trouble?

bioimageio/spec/shared/utils.py
```diff
@@ -182,8 +181,10 @@ def transform_ImportableModule(node):
 
     @staticmethod
     def transform_ResolvedImportableSourceFile(node: ResolvedImportableSourceFile) -> nodes.ImportedSource:
+        module_path = resolve_uri(node.source_file)
+        module_name = os.path.splitext(os.path.split(module_path)[1])[0]
```
FynnBe, line 185: as we predominantly use pathlib, I'd suggest
```suggestion
        module_name = module_path.stem
```
however I don't see why using `module_name` is a good idea here... (see comment below)

bioimageio/spec/shared/utils.py
FynnBe: if repeated import is an issue, we should maybe use ['uuid.uuid5()`](https://docs.python.org/3/library/uuid.html#uuid.uuid5) ?

FynnBe (changes requested):

constantinpape:
Using `f"user_imports.{uuid.uuid4().hex}", resolve_uri(node.source_file)` as module name causes an exception when loading the saved weights again via `torch.jit.load`:
```
Unknown type name '__torch__.user_imports':                                                                                                                                                                                                                   
__torch__.user_imports.0626932a014d4c4e8412d4f96330b42f.UNet2d                                                                                                                                                                                                
~~~~~~~~~~~~~~~~~~~~~~ <--- HERE        
```

I don't know what the issue here within torchscript; it might either be some name-clash with `user_imports` or a problem with the module name being too long. Change to the module name fixes this issue.

FynnBe:
I see... maybe we add a constant prefix to the module name to avoid name clashes? something like `_module_from_source` or `_user_module` or something?

constantinpape:
> I see... maybe we add a constant prefix to the module name to avoid name clashes? something like `_module_from_source` or `_user_module` or something?

Sounds good. I will check later if this solution works and change the code if it does.

constantinpape:
@constantinpape I added `module_from_source` as prefix and this works.

FynnBe (approved):

PR #191, opened 2021-07-20
Rdf modules

FynnBe:
code structure changes as proposed in https://github.com/bioimage-io/spec-bioimage-io/pull/161#issuecomment-876567568


FynnBe:
todo:
 - [ ] use delegated functions in tests

bioimageio/spec/delegator.py
```diff
@@ -0,0 +1,133 @@
+from __future__ import annotations
```
constantinpape, line 1: why is this file called delegator?

constantinpape (commented):
The diff is a bit to large to properly review rn.
But I think it overall looks good; I find the name "delegator" a bit non-descript thoguh.
Also, are there any significant functions renamed or move such that importing them changes?

bioimageio/spec/delegator.py
FynnBe: it delegates the io function to the respective submodule

FynnBe:
> Also, are there any significant functions renamed or move such that importing them changes?

everything importet from `bioimageio.spec` should work as before, but imports from previous submodules are broken.

bioimageio/spec/delegator.py
FynnBe: of course 'utils' or similar would also be appropriate, but there are already many files called 'utils.py' in here...

constantinpape: Is this imported in the top-level, or do library users need to do `from spec.delegator import something`?

constantinpape:
> everything importet from `bioimageio.spec` should work as before, but imports from previous submodules are broken.

Ok, I think that's good enough for now.

PR #192, opened 2021-07-22
substitute stdnum.iso7064.mod_11_2.is_valid

FynnBe:
~~note: not on master but on #191~~

### DIFF
--- a/bioimageio/spec/rdf/v0_2/schema.py
+++ b/bioimageio/spec/rdf/v0_2/schema.py
@@ -1,11 +1,11 @@
 import warnings
 
-import stdnum.iso7064.mod_11_2  # todo: remove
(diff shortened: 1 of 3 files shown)


k-dominik (approved):
lgtm, was there a reason to get rid of this dep other than that we can?

FynnBe:
> lgtm, was there a reason to get rid of this dep other than that we can?

reducing dependencies to make it easier to run in pyodide

k-dominik:
> > lgtm, was there a reason to get rid of this dep other than that we can?
> 
> reducing dependencies to make it easier to run in pyodide

okay less deps are better deps ;) This is always nice! Although that one is Python only, so not much of an issue for pyodide.

FynnBe:
true, it was just about overall package size to speed things up...

PR #193, opened 2021-07-23
Add float32 requirement to preprocessing

constantinpape:
@k-dominik @FynnBe do you think this is sufficient?


k-dominik (approved):
Sure, this is fine. Nitbit: "Should" leaves _some_ space for exceptions. Well, coming from the non-native speaker I am...

constantinpape:
> "Should" leaves _some_ space for exceptions.

Good point, it should be must, also following RFC conventions.

PR #196, opened 2021-07-23
Resource package IO

FynnBe:


k-dominik (commented):
Are those part of the public facing API? Then it would be awesome to see some docstrings. Spontaneously I'm not sure when I would use those, what `package` is in this context (reading the source, I assume a zip archive of some sort - is this specific to models (I suppose it isn't)). Also when would I want to use the _raw_node_ vs the _non-raw-node_ version of the functions.

FynnBe:
> Are those part of the public facing API? Then it would be awesome to see some docstrings. Spontaneously I'm not sure when I would use those, what `package` is in this context (reading the source, I assume a zip archive of some sort - is this specific to models (I suppose it isn't)). Also when would I want to use the _raw_node_ vs the _non-raw-node_ version of the functions.

Yes, I added some doc stirings. 
Currently we only package models, but that might change. All IO is supposed to be somewhat agnostic to RDF type (and version).

FynnBe:
maybe `load_node` and `load_raw_node` should just also accept file path/URLs to packages? we could just decide based on extension? (.yaml vs .zip) and then absorb `import_package`...

oeway:
> maybe `load_node` and `load_raw_node` should just also accept file path/URLs to packages? we could just decide based on extension? (.yaml vs .zip) and then absorb `import_package`...

I think it make sense to combine all these.

For the `node` and `raw_node`, can we discuss a bit on an alternative name for `node` which is more intuitive? The `node` sounds like we are dealing with a tree (is that because of the yaml structure is like a tree?). 

I might misunderstand the concept, but after loading a model RDF, shouldn't we get a sort of Model instance which has a bunch of attributes and functions? Then if it's `raw_node`, can we call it something like `ModelConfig`?

constantinpape:
> maybe `load_node` and `load_raw_node` should just also accept file path/URLs to packages?

I agree, I would also overload it such that this supports both rdfs and zipped folders.

> For the `node` and `raw_node`, can we discuss a bit on an alternative name for `node` which is more intuitive? The `node` sounds like we are dealing with a tree (is that because of the yaml structure is like a tree?).
> 
> I might misunderstand the concept, but after loading a model RDF, shouldn't we get a sort of Model instance which has a bunch of attributes and functions? Then if it's `raw_node`, can we call it something like `ModelConfig`?

I think `Node` comes from marshmallow?! I also don't find it intuitive here.

FynnBe:
'node' comes from the tree. I totally agree that it's not intuitive. It's too general. Within the code it makes some sense, but in the exposed functions it should carry different names, maybe `resource` as it comes form the resource description file?

`raw_resource` and `resource` or `evaluated_resource`?
The naming of raw vs ' '/'evaluated' could also do with an update..
- 'raw': content from the RDF yaml but as annotated dataclass structure with some more pythonic types (paths are pathlib.Paths not str, etc)
- '' ('evaluated'): enriched with convenience functionality, e.g. imported source code, urls are resolved to local (cached) pahths, etc..

FynnBe:
> I might misunderstand the concept, but after loading a model RDF, shouldn't we get a sort of Model instance which has a bunch of attributes and functions? Then if it's raw_node, can we call it something like ModelConfig?

to clear this up (using the current terms): 
- a `model RDF` can be loaded with `load_raw_node` to yield a pythonic representation of its raw content. (with full type annotations and based on dataclasses)
- a `model RDF` or a `raw_node` can be loaded with `load_node` to do the above and transform the `raw_node` with a fixed set of `NodeTransformer`s: URLs are downloaded and replaced by local cached paths ('resolved'), source code is imported and available as a callable [ImportedSource](https://github.com/bioimage-io/spec-bioimage-io/blob/4c6b467c62d9d6ccd0e059156e8aca71607ac276/bioimageio/spec/shared/nodes.py#L23-L28) node.

transformations from raw node to node: 
https://github.com/bioimage-io/spec-bioimage-io/blob/4c6b467c62d9d6ccd0e059156e8aca71607ac276/bioimageio/spec/shared/utils.py#L343-L348
(As some nodes might not be transformed, some resolved nodes are identical to raw nodes.)

As we look at the model information as a tree I find the node term acceptable in the code, but it is not useful to expose. Thus my above suggestion to use `resource` instead...  But I would be totally fine with changing it to `config.Model` and `raw_config.Model` instead or also `raw_config.ModelRawConfig` (or maybe `raw_config.RawModelConfig`?) and `config.ModelConfig`. We used to avoid 'config' as we used to refer to the model RDF as a model config and we needed something distinct. But in our current situation I like 'config' more and more... 

```python
def resolve_raw_config_to_config(raw_config: GenericRawConfig, root_path: os.PathLike, configs_module: ModuleType) -> GenericConfig: 
     """resolve all uris and sources""" 
     config = UriConfigTransformer(root_path=root_path).transform(raw_config) 
     config = SourceConfigTransformer().transform(config) 
     config = RawConfigTypeTransformer(configs_module).transform(config) 
     return config 

``` 
    
Additional functions could be added to the (resolved) node objects (not the raw nodes!), but currently they do not implement anything.

constantinpape:
I think renaming `node` to `resource` in the user facing functions is a good start. I would not call it anything with `model`, because it can represent both models or other types of rdfs (at least that's how I understand it). We could also call it `resource_description`, to map more closely to `rdf`.

Also, I think `resolve_raw_x_to_x` should just be called `resolve_raw_x`.

FynnBe:
> We could also call it resource_description, to map more closely to rdf.

how about renaming 'package' to 'resource_package' to emphasize their analogy?

constantinpape:
> how about renaming 'package' to 'resource_package' to emphasize their analogy?

Do you mean, for example `import_package` -> `import_resource_package`? I think that's a good idea, also because package is a bit too generic for my taste.

FynnBe:
> Do you mean, for example import_package -> import_resource_package? I think that's a good idea, also because package is a bit too generic for my taste.

yes, exactly!

FynnBe:
after merging #204 this PR should be ready to be merged as well...

bioimageio/spec/shared/utils.py
```diff
@@ -340,45 +347,14 @@ def _download_uri_to_local_path(uri: typing.Union[nodes.URI, raw_nodes.URI]) ->
     return local_path
 
 
-def resolve_raw_node_to_node(raw_node: GenericRawNode, root_path: os.PathLike, nodes_module: ModuleType) -> GenericNode:
+def resolve_raw_node(raw_node: GenericRawNode, root_path: os.PathLike, nodes_module: ModuleType) -> GenericNode:
```
constantinpape, line 350: Is this user facing? If it is we should maybe rename to `resolve_raw_resource_description`?

constantinpape (approved):
Maybe rename `resolve_raw_node`? Looks good otherwise.

bioimageio/spec/shared/utils.py
FynnBe: this is not user facing and only used in shared.io_ where the node/raw_node names are still used. I'm working on a separate PR to improve the node raw_node distinction and will change the names in io_ there as well. But all of this is not user facing...

PR #199, opened 2021-07-26
Overload load to accept zip files

FynnBe:


constantinpape (commented):
So with this `load_node` / `load_raw_node` excepts both the path to the rdf or the zipped folder?
Can you please also add a test that checks that it actually works?

bioimageio/spec/model/v0_1/raw_nodes.py
```diff
@@ -19,6 +19,9 @@
 FormatVersion = Literal["0.1.0"]
 
 
+Dependencies = Dependencies
```
constantinpape, line 22: ???

bioimageio/spec/model/v0_3/raw_nodes.py
```diff
@@ -28,6 +28,7 @@
 # same as general RDF
 Badge = Badge
 CiteEntry = CiteEntry
+Dependencies = Dependencies
```
constantinpape, line 31: Ah, I guess this is a common pattern :D. But it looks weird enough that a comment is warranted.

constantinpape (approved):
Some minor comments, looks good overall.

bioimageio/spec/model/v0_3/raw_nodes.py
FynnBe: yeah... I don't know what the way to go is there...  just importing it and not using it looks like an unrequired import.

constantinpape: yeah, maybe just write a comment to explain it

FynnBe: added comments in 44f4d5e

PR #205, opened 2021-07-29
Improve nodes

FynnBe:
this PR makes `raw_nodes` distinct from `nodes` by adding `base_nodes`. These changes are internal only and should not affect the use of the library, other than returned types (that behave the same as before though).

`raw_nodes.RawNode` and `nodes.Node` now inherite from `base_nodes.NodeBase` (before: Node= RawNode):
https://github.com/bioimage-io/spec-bioimage-io/blob/b07a537242607cdeb0eea6782921b1697208c137/tests/test_base_nodes.py#L1-L11

This allows for better type checking (more files are now checked with mypy).

`base_nodes` are abstract, whenever specifying properties that differ between `raw_nodes` and `nodes`.
Although abstract dataclasses are fine with python, mypy has an issue for which a workaround is used like suggested in https://github.com/python/mypy/issues/5374#issuecomment-650656381 (this is commented where used), e.g.:
https://github.com/bioimage-io/spec-bioimage-io/blob/b07a537242607cdeb0eea6782921b1697208c137/bioimageio/spec/rdf/v0_2/base_nodes.py#L97-L110

To ensure `base_nodes` are not instantiated directly an additional check was added:
https://github.com/bioimage-io/spec-bioimage-io/blob/b07a537242607cdeb0eea6782921b1697208c137/bioimageio/spec/shared/base_nodes.py#L27-L33


FynnBe:
currently still a minor problem:
when initializing a `raw_nodes.RDF()`, the base_nodes.NodeBaser.__post_init__() is called, but no dataclass fields are registered at this point... (this might be by design?! If so the check for missing args by type needs to move... )

bioimageio/spec/model/v0_3/base_nodes.py
```diff
@@ -0,0 +1,211 @@
...
+    weights_format_name = "Tensorflow Saved Model"
+    tensorflow_version: Union[_Missing, distutils.version.StrictVersion] = missing
+    # tag: Optional[str]  # todo: check schema. only valid for tensorflow_saved_model_bundle format
+
+
+class TensorflowSavedModelBundleWeightsEntry(_TensorflowSavedModelBundleWeightsEntry, ABC):
+    pass
+
+
+# WeightsEntry = Union[
```
constantinpape, line 151: ???

constantinpape (commented):
Looks good overall; just one minor comment.

PR #206, opened 2021-07-30
a closer look at URI

FynnBe:


constantinpape (approved):

FynnBe:
properly annotating the URI nodes is a bit tricky, but I think this checks out now...

constantinpape:
Ok, everything is passing now, so I'm merging it.

Issue #209, opened 2021-07-30
replace ruamel.yaml with vanilla pyyaml


FynnBe:
for the sake of simplicity for pyiodide

@oeway  @emilmelnikov

PR #212, opened 2021-08-02
Remove ruamel.yaml dependency

FynnBe:
ruamel.yaml is not supported by Pyodide, therefor we use PyYAML instead.

problem: PyYAML currently only supports yaml 1.1, whereas ruamel.yaml (what we used until now) defaults to yaml 1.2. 
Differences: https://yaml.readthedocs.io/en/latest/pyyaml.html#defaulting-to-yaml-1-2-support

The one difference that might be most relevant for us: "correct parsing of floating point scalars with exponentials" is added manually to the current ruamel.yaml replacement.
see https://github.com/yaml/pyyaml/issues/173 and https://stackoverflow.com/questions/30458977/yaml-loads-5e-6-as-string-and-not-a-number.

Also note that not all valid json is compatible with yaml 1.1 (but with yaml 1.2).

With this yaml 1.1/1.2 hybrid we should be good for now and can hopefully switch easily to the full yaml 1.2 in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This repository contains specifications defined by the BioImage.IO community. Th
 
 
 All the BioImage.IO-compatible RDF must fulfill the following rules:
- * Must be a YAML file encoded as UTF-8;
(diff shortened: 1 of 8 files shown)


FynnBe:
fixes #209

constantinpape (approved):
LGTM, feel free to merge.

oeway:
@FynnBe Great! Could you make a new release so we can try it with pyodide?

FynnBe:
sure, let's wait for https://github.com/bioimage-io/spec-bioimage-io/pull/214 to be merged and then go ahead with a little new release

PR #217, opened 2021-08-03
Update __main__.py

constantinpape:
Do we need to keep this around? We don't have many users around so I would vote for removing deprecated functionality fast.


FynnBe (approved):
Yep, let's get rid of it!

PR #226, opened 2021-08-05
Make bioimageio.spec purely about the spec

FynnBe:


constantinpape:
That's a big diff...
Could you summarize:
- which functionality stays here?
- which functionality would go to `bioimageio.core`?

Also, why are you removing the exmaple model spec?

FynnBe:
> Also, why are you removing the exmaple model spec?

I'm thinking it shoulds stay here.. my reasoning was that it requires io which will only be available in the tests, so the package wouldn't be able to work with the example... but now I'm thinking we can add a simple yaml reader, use it only in the command, and throw a runtime exception if it's not available... 

> which functionality stays here?
- definition of spec: schema, raw_nodes, nodes
- definition of package: `get_package_content`: change a raw resource def to local files and return names of local files with their remote locations (but does not resolve remote locations)
- utils: transformers raw_node <-> nodes (without `resolve_uri`? to avoid io)
- input expected as dict (single exception: cli can read yaml)

moving out (to bioimageio.core):
- `resolve_uri`, accepting RDFs as files 
- make an actual RD package (using `get_package_content`)

constantinpape (approved):
In principle this looks good, but I don't wanna merge until https://github.com/bioimage-io/python-bioimage-io/pull/93 can also be merged.

constantinpape:
@FynnBe is this ready to be merged now? If so, could you please summarize what has changed now, especially what user facing functionality is still in this package? (The diff is too massive for me to look at now).

FynnBe:
> @FynnBe is this ready to be merged now? If so, could you please summarize what has changed now, especially what user facing functionality is still in this package? (The diff is too massive for me to look at now).

sure thing. functionality remaining here is:
- get_resource_package_content  (excluding the 'rdf.yaml' itself, which requires yaml to serialize to string...) <- should we maybe rename this also to avoid name clash with this function in core that does include the rdf.yaml...? yeah.. I think so... 
- load_raw_resource_description (only accepts a dict, not path, url, etc... has a doppelgänger in core that does accpet paths and urls)  <- rename to `load_raw_resource_description_from_dict`?
- serialize_raw_resource_description_to_dict

constantinpape:
> sure thing. functionality remaining here is:
> 
>     * get_resource_package_content  (excluding the 'rdf.yaml' itself, which requires yaml to serialize to string...) <- should we maybe rename this also to avoid name clash with this function in core that does include the rdf.yaml...? yeah.. I think so...
> 

What exactly does `get_resource_package_content` do? What exactly do you mean by excluding the `rdf.yaml`?

>     * load_raw_resource_description (only accepts a dict, not path, url, etc... has a doppelgänger in core that does accpet paths and urls)  <- rename to `load_raw_resource_description_from_dict`?
> 

Yes, I would rename it.

>     * serialize_raw_resource_description_to_dict

FynnBe:
todo:
 - [ ] rename load_raw_resource_description to load_raw_resource_description_from_dict

constantinpape:
@FynnBe maybe I am misremembering something, but didn't we decide to move `load_raw_resource_description` and `serialize_raw_resource_description` back here, because we will need the former for the spec validation function (which should stay in this repository)? And just throw an exception if one of the functions is used without having `yaml`?!

FynnBe:
> @FynnBe maybe I am misremembering something, but didn't we decide to move `load_raw_resource_description` and `serialize_raw_resource_description` back here, because we will need the former for the spec validation function (which should stay in this repository)? And just throw an exception if one of the functions is used without having `yaml`?!

we talked about `serialize...` and `save...`, but following our logic for those I see that `load_raw...` also falls in this category.. I'll move that one, too then. (allow paths and uris and not only dicts as it is now..)

FynnBe:
turns out, if we allow to load from urls as well this moves quite some functionality over here (including the caching). Maybe we should only allow paths and keep the url support in core?

constantinpape:
> turns out, if we allow to load from urls as well this moves quite some functionality over here (including the caching). Maybe we should only allow paths and keep the url support in core?

I think it's fine to have URL support only in core. I think the minimal functionality in spec is to provide a validation function in python when passing a dict and via the command line when passing a file path (which needs yaml as additional functionality as discussed). All the rest can go to core.

FynnBe:
> I think it's fine to have URL support only in core. I think the minimal functionality in spec is to provide a validation function in python when passing a dict and via the command line when passing a file path (which needs yaml as additional functionality as discussed). All the rest can go to core.

the command even has rudimentary url support (for validation only) and the library function now support for paths with yaml (but no urls). so this should cover all regular use cases. and for more users would need to switch to bioimageio.core

PR #231, opened 2021-09-09
Update VERSION

constantinpape:
I think we should make a new release, now that @FynnBe has moved the more complex functionality to `bioimageio-python`

### DIFF
--- a/bioimageio/spec/VERSION
+++ b/bioimageio/spec/VERSION
@@ -1,3 +1,3 @@
 {
-    "version": "0.3.2post2"
+    "version": "0.3.2post3"
 }


k-dominik:
are we sure that bioimage.core PRs will not make any changes here necessary? We have tests there, that test against main branch here, and that still fail.

constantinpape:
I think the tests vs. `main` are actually passing.

k-dominik (approved):
cool let's go ahead

constantinpape:
> cool let's go ahead

Awesome! 
@FynnBe please go ahead and merge this if there are no objections from your side (i.e. nothing left in the refactoring that I forgot about)

PR #233, opened 2021-09-12
More examples

constantinpape:
Add example models for:
- output shape that is different from the input shape
- fixed shape
- multiple input / output tensors


FynnBe (approved):
Awesome additional examples! 👍

FynnBe:
we should make use of these examples in python-bioimage-io as well

constantinpape:
> we should make use of these examples in python-bioimage-io as well

Yes, I added these so we have test cases for multi tensor input and explicit output shape.

PR #235, opened 2021-09-14
ImplicitShape for inputs and outputs (part 1 in 0.3)

FynnBe:
This PR changes implements the first set of changes discussed in #234 

also it:
- rename `ImplicitInputShape` to `ParametrizedInputShape` to avoid confusion

PR #236, opened 2021-09-15
No duplicate tensor names (in v0.4)

FynnBe:
in order to allow `ImplicitShape` (with its `.reference_tensor` attribute) to be used across inputs and outputs alike this PR add the restriction for any tensor name not to be a duplicate.

As this is also the first real time we are bumping the minor version, please look at this PR also with respect to how we want to progress to following minor versions (copy paste vs import from earlier version, etc).

note: builds ontop of #235


FynnBe:
closed in favor of #297

PR #237, opened 2021-09-15
bump version to 0.3.3

FynnBe:
just making sure: it's vanilla 0.3.3 without any post, correct?


constantinpape (approved):
Looks good, let's merge when all tests pass.

FynnBe:
do we keep these bump version branches around or do we remove them?
as the tags are there for reference, I'd suggest to remove the (inconsistently named) branches...

constantinpape:
No need to keep the branches.

PR #239, opened 2021-09-16
polish up pre- and postprocessing tests and schemas

FynnBe:


constantinpape (approved):
I don't have time for a proper review of the test changes. But the changes in schema look good, and at a quick glance it also looks like pre/and-postprocessing in schema and the markdown file agree.
(We should eventually generate the markdown from the schema, but let's do this after the release.)

Issue #242, opened 2021-09-22
Verify a yaml string


oeway:
@FynnBe I am looking for a way to verify a yaml string in the bioengine, I confirm that I can import bioimageio.spec, could you provide a short script for me to verify a yaml string?

FynnBe:
We didn't expect that use case actually, but I made a PR to support it: https://github.com/bioimage-io/spec-bioimage-io/pull/243

This is because calling the validate command with a yaml string is somewhere in between calling it on a file (url or path) and calling it with a dictionary (what you would get if you load a file with yaml). 

Calling it with a yaml string obviously still has the ruamel.yaml dependency. If you want to avoid it, you need to pass the string beforehand and call the validate function with a dictionary instead. 

```python
from bioimageio.spec.commands import validate

from ruamel.yaml import YAML

yaml = YAML(typ="safe")

# preload yaml 
source = yaml.load("spec_file_or_string")
# no yaml dependency below, unless source is not a dict, but instead:
# source = "existing_spec_source_path"  # or as os.PathLike, e.g. pathlib.Path
# or
# source = "https_spec_source_url"
# or  with PR #243
# source= "name: 'yaml string model'"

code = validate(source)
if code == 0:
    print("valid")
else:
    print("invalid")
```

FynnBe:
note that pyyaml uses YAML 1.1 (not YAML 1.2 like ruamel.yaml) and pyyaml would also require `yaml.safe_load("yaml_string")` to load a string instead of a file, but we should not use pyyaml as valid specs need to be written in YAML 1.2 so in python we should stick to ruamel.yaml.

constantinpape:
> If you want to avoid it, you need to pass the string beforehand and call the validate function with a dictionary instead.

I think that would be the better option in the browser:
- load the dict representation with yaml in js
- pass the dict to python (can be json encoded)
- validate dict in python

This way you don't need the yaml dependency.

oeway:
Both works for me, a string or a dictionary, go ahead and implement any you think the best.
Thank you!

FynnBe:
dictionary has been implemented first and #243 now adds support for the yaml strings. In conjuntion with pyodide you should opt for the dictionary as @constantinpape sketched aboved, which already works now.

oeway:
Ok, great!

@FynnBe In your example use the valid function, how do I get the detailed error message if it's not valid?

FynnBe:
[...]

PR #243, opened 2021-09-22
support loading yaml from str

FynnBe:
#242 got me looking into validating a yaml string, which we do not currently support. This PR changes that.

Until now, validate excepts a string, but only tries to interpret it as a URL (if it starts with 'https' or as a path (if that path exists)). This PR tries to load the string with ruamel.yaml if interpretation as URL and path failed.

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -36,11 +36,9 @@ def validate(
 
             if is_path:
                 rdf_source = pathlib.Path(rdf_source)
-            else:
-                raise RuntimeError(f"Could not retrieve {rdf_source}")
 
         if yaml is None:
-            raise RuntimeError("Cannot validate from file without ruamel.yaml dependency!")
+            raise RuntimeError("Cannot validate from file or yaml string without ruamel.yaml dependency!")
 
         rdf_source = yaml.load(rdf_source)
 


bioimageio/spec/commands.py
```diff
@@ -36,11 +36,9 @@ def validate(
...
             if is_path:
                 rdf_source = pathlib.Path(rdf_source)
-            else:
-                raise RuntimeError(f"Could not retrieve {rdf_source}")
 
         if yaml is None:
-            raise RuntimeError("Cannot validate from file without ruamel.yaml dependency!")
+            raise RuntimeError("Cannot validate from file or yaml string without ruamel.yaml dependency!")
 
         rdf_source = yaml.load(rdf_source)
```
constantinpape, line 43: Does this really work? Usually the syntax would be `yaml.loads(rdf_source)` if this is a string encoded dict.
(I am not so sure about yaml though, but this is the python default way of doing it, e.g. for pickle or json)

FynnBe: `yaml.load(rdf_source)` only works for ruamel.yaml's `yaml=YAML(typ="safe")` instance, but since that's what we are using it should be good to go...

FynnBe: ... (given that rdf_source) is a yaml encoded string

constantinpape: Ok, guess ruamel choose a bit of a weird API

constantinpape (approved):
LGTM, feel free to merge whenever this is ready @FynnBe

PR #244, opened 2021-09-22
fix kwargs annotations in raw_nodes

FynnBe:
`raw_nodes.Preprocessing.kwargs`  and `raw_nodes.Postprocessing.kwargs` missed the `_Missing` annotation, thus alwyas expecting a dict, but the schema (and spec) keep `kwargs` optional.


k-dominik (approved):
ah right :) given the fix, the error message makes more sense ;)

constantinpape (approved):

PR #245, opened 2021-09-23
Hpa model

constantinpape:
Adds hpa densenet model for image classification, example provided by @oeway.

@oeway I have adapted the model slightly to test different values per channel in `scale_linear`.

@FynnBe this model currently fails the spec, but I think it should be valid.
We provide a value per channel here and set axes to `xy` (the image has 4 channels).
So I think there is a bug in the validator, see error message below. Could you please look into this?

```
self = <ScaleLinear(many=False)>
data = {'axes': 'xy', 'gain': array([0.00392157, 0.00392157, 0.00392157, 0.00392157]), 'offset': array([0., 0., 0., 0.])}
kwargs = {'many': False, 'partial': False}

    @validates_schema
    def either_gain_or_offset(self, data, **kwargs):
>       if data["gain"] == 1.0 and data["offset"] == 0:
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```


constantinpape (commented):
Thanks for fixing the issue, @FynnBe!

oeway:
@constantinpape @FynnBe This is awesome! Thanks a lot!

PR #248, opened 2021-09-27
remove SpecURI from model v0.1

FynnBe:
The spec url functionality for nested RDFs in model RDF 0.1 was never fully supported. This PR turns those special URIs into regular `field.URI`s


constantinpape (approved):
I don't really understand this, but I like code removal :D.
So I trust that this is doing the right thing here.

k-dominik (approved):
I guess this was this idea from the beginning that transformations and other stuff could come from git?!

In any case, I'm with @constantinpape :D

FynnBe:
yes, spot on

PR #249, opened 2021-09-28
force encoding to utf-8

k-dominik:
read_text takes the default locale... Now fun fact - when packaging ilastik for OSX and starting via double click, this seems to default to ascii

### DIFF
--- a/bioimageio/spec/VERSION
+++ b/bioimageio/spec/VERSION
@@ -1,3 +1,3 @@
 {
-    "version": "0.3.3post1"
+    "version": "0.3.3post2"
 }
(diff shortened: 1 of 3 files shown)


FynnBe (approved):
nice catch!

should we also update setup.py?

k-dominik:
better safe than sorry ;) Who knows what kind of locales people set...

k-dominik:
should I go ahead and merge this? Also could we make a `post2` release with it (before Friday)?

oeway (approved):
Looks good to me.

Go ahead and change the version to post2, so we release it already when it's merged.

k-dominik:
cool - as it turned out the issue fixed in this PR would also occur on other OSes if with non-utf8 locales... (also discovered _during_ a course :dagger: )

k-dominik:
allright, I'll go ahead an merge this! thx @oeway and @FynnBe for the review :)

PR #250, opened 2021-10-05
read "rdf.yaml" from zipfile if source is zipped

FynnBe:
also adds support for BytesIO, StringIO and bytes as rdf_source

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -1,7 +1,9 @@
 import os
 import pathlib
 import traceback
-from typing import Any, Dict, IO, List, Sequence, Union
+import zipfile
+from io import BytesIO, StringIO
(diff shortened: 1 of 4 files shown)


constantinpape (commented):
~~I don't really like that we are duplicating all the code for `validate` in `spec-bioimage-io` and `core-bioimage-io`. Can we refactor this a bit to reuse some of the code?~~

Nevermind the comment above, I misunderstood the PR. Are these changes already part of #251?

FynnBe:
> Nevermind the comment above, I misunderstood the PR. Are these changes already part of #251? 

yes I merged this one intot #251 to avoid merge conflicts. Though we would merge this one before and thus slim down the doi one (to keep PR sizes small...), but now discussion about these additions have started in #251 , so I'll close this one

PR #251, opened 2021-10-05
add doi support for validate command

FynnBe:
todos:
- [ ] add test

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -1,12 +1,13 @@
 import os
 import pathlib
+import re
 import traceback
 from typing import Any, Dict, IO, List, Sequence, Union
 
 from marshmallow import ValidationError
 
 from .io_ import load_raw_resource_description
-from .shared import yaml
+from .shared.common import DOI_REGEX, yaml
 
 
 def validate(
@@ -22,6 +23,9 @@ def validate(
     source_name = str(rdf_source.get("name") if isinstance(rdf_source, dict) else rdf_source)
     if not isinstance(rdf_source, dict):
         if isinstance(rdf_source, str):
+            if re.fullmatch(DOI_REGEX, rdf_source):  # turn doi into url
+                rdf_source = f"https://doi.org/{rdf_source}"
+
             if rdf_source.startswith("http"):
                 from urllib.request import urlretrieve
 
--- a/bioimageio/spec/shared/common.py
+++ b/bioimageio/spec/shared/common.py
@@ -15,6 +15,9 @@ else:
     yaml = YAML(typ="safe")
 
 
+DOI_REGEX = r"^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$"
+
+
 def get_format_version_module(type_: str, format_version: str):
     assert "." in format_version
     import bioimageio.spec


FynnBe:
current problem: I don't have a working regex (in python)

Great, I found this doi regex: `^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$` [here](https://www.crossref.org/blog/dois-and-matching-regular-expressions/) (the first one for the "modern" dois)

Great it works with this example doi: `10.5072/zenodo.886788`
[This](https://regex101.com/) online regex tool tells me it matches:
image

BUT:
```python
>>> import re
>>> print(re.fullmatch("^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$", "10.5072/zenodo.886788"))
>>> None
>>> print(re.match("^10.\d{4,9}\/[-._;()\/:A-Z0-9]+$", "10.5072/zenodo.886788"))
>>> None
```

HELP!

bioimageio/spec/commands.py
```diff
@@ -19,9 +22,26 @@ def validate(
...
 
-    source_name = str(rdf_source.get("name") if isinstance(rdf_source, dict) else rdf_source)
+    if isinstance(rdf_source, (BytesIO, StringIO)):
+        rdf_source = rdf_source.read()
+    elif isinstance(rdf_source, os.PathLike):
+        # zipfile.is_zipfile cannot deal with all os.PathLike objects (e.g. pytest's `LocalPath`),
+        # so we cast to pathlib.Path
+        rdf_source = pathlib.Path(rdf_source)
+
+    source_name = str(
```
constantinpape, line 32: Why are we doing this before loading the rdf as a dict in case it's online?

bioimageio/spec/commands.py
```diff
@@ -19,9 +22,26 @@ def validate(
     if update_format_inner is None:
         update_format_inner = update_format
 
-    source_name = str(rdf_source.get("name") if isinstance(rdf_source, dict) else rdf_source)
+    if isinstance(rdf_source, (BytesIO, StringIO)):
```
constantinpape, line 25: I don't fully understand what's going on here, but in general this code would be easier to understand if it would look like
```python
if isinstance(rdf_source, (BytesIO, StringIO)):
   ...
elif isinstance(rdf_source, os.PathLike)
   ...
elif isinstance(rdf_source, str) and rdf_source.startswith("http):
   ...
elif isinstance(rdf_source, str) and re.fullmatch(DOI_REGEX, rdf_source):   
   ...
# some more options I overlooked?   
else:
   if not isinstance(rdf_source): raise ValueError   
```

constantinpape (commented):
Left some code comments.
In general, should we refactor the code that parses the rdf_source into some helper function so that it can be used in `load_resource_description` etc. as well?!

bioimageio/spec/commands.py
FynnBe: `source_name` is only used for more concise error message

bioimageio/spec/commands.py
FynnBe: 👍 yeah.. that grew a bit over time tbh...

FynnBe:
> In general, should we refactor the code that parses the rdf_source into some helper function so that it can be used in load_resource_description etc. as well?!

yes, we should 👍

PR #257, opened 2021-10-15
improve summary output of validate

FynnBe:
looking at how the `validate` command is now used on model upload, I thought it could be enhanced by making it a summary. 

The summary is a dict of the form:
```python
{
        "name": source_name if raw_rd is None else raw_rd.name,
        "source_name": source_name,
        "error": error,
        "traceback": tb,
        "nested_errors": nested_errors,
    }
```

Pros:
 - more explicit information
 - output just looks better
 - easy extension with 'warnings' key (wherever that is implemented)

Cons:
 - change from before: The returned dictionary now will never be empty. Need to check if the 'error' key is `None` 
 

 (note: this does not change the error message for the Union field)

### DIFF
--- a/bioimageio/spec/__main__.py
+++ b/bioimageio/spec/__main__.py
@@ -19,15 +19,18 @@ def validate(
     update_format_inner: bool = typer.Option(
         None, help="For collection RDFs only. Defaults to value of 'update-format'."
     ),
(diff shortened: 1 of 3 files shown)


constantinpape (approved):
LGTM, but let's rerun the tests before merging.

oeway:
Looks good! Please also update the version to make a new release so I can use it on the website.

Issue #259, opened 2021-10-18
verify model config files - ruamel.yaml dependency


esgomezm:
Hi!

I'm not sure of this part of the validator is supported but I get the following error message [here](https://github.com/deepimagej/models/runs/3930902817#step:7:58):

RuntimeError: Cannot read RDF from manifest.bioimage.ya... without ruamel.yaml dependency!
Error: Process completed with exit code 1.

We are uding this validator as a [github workflow for our models](https://github.com/deepimagej/models/blob/master/.github/workflows/compile-manifest.yml#L41). 

Thank you!

constantinpape:
You need to install ruamel.yaml as additional dependency. see https://pypi.org/project/ruamel.yaml/.
Note that this is deliberately left out of the `requirements.txt` in order to enable running `bioimageio.spec` in pyodide.

esgomezm:
I added the dependency but still it outputs an error message that I don't know how to solve it: 

```
Run pip install git+https://github.com/bioimage-io/spec-bioimage-io.git
Collecting git+https://github.com/bioimage-io/spec-bioimage-io.git
  Cloning https://github.com/bioimage-io/spec-bioimage-io.git to /tmp/pip-req-build-u79dpdi9
  Running command git clone --filter=blob:none -q https://github.com/bioimage-io/spec-bioimage-io.git /tmp/pip-req-build-u79dpdi9
  Resolved https://github.com/bioimage-io/spec-bioimage-io.git to commit 8f5396652723a462d604a4f57ba5fdaefc7ce10d
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing wheel metadata (pyproject.toml): started
  Preparing wheel metadata (pyproject.toml): finished with status 'done'
Requirement already satisfied: marshmallow-jsonschema in /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages (from bioimageio.spec==0.3.3.post4) (0.12.0)
Requirement already satisfied: typer in /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages (from bioimageio.spec==0.3.3.post4) (0.4.0)
[...]

PR #261, opened 2021-10-19
Look for more rdfs

FynnBe:
builds on top of #260

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -6,6 +6,7 @@ from typing import Dict, IO, Optional, Union
 from marshmallow import ValidationError
 
 from .io_ import load_raw_resource_description, resolve_rdf_source
+from .shared.common import nested_default_dict_as_nested_dict
 
 KNOWN_COLLECTION_CATEGORIES = ("application", "collection", "dataset", "model", "notebook")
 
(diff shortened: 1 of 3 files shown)


constantinpape (commented):
I don't have a strong opinion on this.
@esgomezm @oeway this is changing the validator so that we allow `rdf.yml`, `rdf.yaml`, `model.yaml` and `model.yml` instead of only `rdf.yaml` as name of the model resource description file in a zipped model. This would probably fix the errors we see in #259, but we could not rely on always having the exact file `rdf.yaml`, so this might supporting modelzoo resources a bit more cumbersome in other tools that don't use `bioimageio.spec`.

FynnBe:
> I don't have a strong opinion on this. @esgomezm @oeway this is changing the validator so that we allow `rdf.yml`, `rdf.yaml`, `model.yaml` and `model.yml` instead of only `rdf.yaml` as name of the model resource description file in a zipped model. This would probably fix the errors we see in #259, but we could not rely on always having the exact file `rdf.yaml`, so this might supporting modelzoo resources a bit more cumbersome in other tools that don't use `bioimageio.spec`.

actually this would currently only change what we expect in a zenodo record not a packaged zip.
But if we do want this I'll add it for packaged zips analogously...

PR #262, opened 2021-10-19
Fail proof validate

FynnBe:
builds on top of #260 

don't fail due to bad (non-existing, etc) sources, but return that failure as an error in the expected summary instead.


constantinpape (approved):
LGTM, @FynnBe please go ahead and merge this and/or #260 how you see fit.

PR #267, opened 2021-10-22
Add keras and tf example model

constantinpape:
Example model with keras / tf weights provided by @esgomezm.
Still WIP because this does not pass all relevant tests, see comments below.


constantinpape:
Closed in favor of #276

PR #268, opened 2021-10-30
Better Union field error message

FynnBe:
This PR adds an explanation to the multiple error lines resulting from a failed Union field:
The new output has the form:
 - 'Errors in all options. Fix any of the following errors:', 
 - <error for option 1>
 - <error for option 2> 
 
The individual errors (which are lists themselves) are sorted by length to more easily find the most straightforward fix.

### DIFF
--- a/bioimageio/spec/shared/fields.py
+++ b/bioimageio/spec/shared/fields.py
@@ -4,8 +4,6 @@ import distutils.version
 import logging
 import pathlib
 import typing
-from urllib.parse import urlparse
-from urllib.request import url2pathname
 
 import marshmallow_union
 import numpy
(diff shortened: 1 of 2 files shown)


bioimageio/spec/shared/fields.py
```diff
@@ -169,6 +167,14 @@ def __init__(self, *super_args, **super_kwargs):
         super().__init__(*super_args, **super_kwargs)
         self.type_name += f"\\[{' | '.join(cf.type_name for cf in self._candidate_fields)}\\]"  # add types of options
 
+    def _deserialize(self, value, attr=None, data=None, **kwargs):
+        try:
+            return super()._deserialize(value, attr=attr, data=data, **kwargs)
+        except ValidationError as e:
+            errors = sorted(e.messages, key=lambda msg: len(msg))
+            messages = ["Errors in all options for this field. Fix any of the following errors:"] + errors
```
constantinpape, line 175: Does this parse well? If I try this the output is just a single list without new lines:
```
raise ValidationError(message=["a", "b", "c"])
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
<ipython-input-3-e0174c71d04f> in <module>
----> 1 raise ValidationError(message=["a", "b", "c"])

ValidationError: ['a', 'b', 'c']
```

constantinpape (approved):
Looks good in general. It would be nice to make this even more readable by printing each error message in a new line, but I am not sure if there is a way to get this with marshmallow validation errors.

bioimageio/spec/shared/fields.py
FynnBe: none of the error messages that are returned anywhere are followed by new lines. This formatting is left to the code displaying the error message. When using the `validate` command pretty pprint is used to display the nested errors. I think we should leave the formatting there

PR #269, opened 2021-11-04
Add --version option to CLI

FynnBe:
displays help on receiving `--version` argument

a separate `--version` option is not straight forward to implement with typer as it complains about a missing command. But making the `--version` option available without command leads to the CLI not complaining about a missing command even without the `--version` option. It is just overall simpler to add it as a help alias, especially because the version is very relevent for the help as well.

To make the verion/help even more useful this PR also adds the format versions of the implemented RDFs (general RDF and model RDF) to the help/version string:

```
Usage: bioimageio [OPTIONS] COMMAND [ARGS]...
```
new help/version string from this PR:
```
  bioimageio.spec 0.3.3post6
  implementing:
          general RDF 0.2.0
          model RDF 0.3.3
```
```
Options:
  --install-completion   Install completion for the current shell.
  --show-completion      Show completion for the current shell, to copy it or
                         customize the installation.

  -h, --version, --help  Show this message and exit.

Commands:
  validate  Validate a BioImage.IO Resource Description File (RDF).

```

### DIFF
--- a/bioimageio/spec/__main__.py
+++ b/bioimageio/spec/__main__.py
@@ -2,11 +2,17 @@ from pprint import pprint
 
 import typer
 
-from bioimageio.spec import __version__, commands
+from bioimageio.spec import __version__, commands, model, rdf
 
-help_version = f"bioimageio.spec package version {__version__}"
+help_version = (
+    f"bioimageio.spec {__version__}"
+    "\nimplementing:"
+    f"\n\tgeneral RDF {rdf.format_version}"
+    f"\n\tmodel RDF {model.format_version}"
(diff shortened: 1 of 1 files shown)


FynnBe:
the help/version string is also shown when running a command (just like before):
```
$bioimageio validate lala
bioimageio.spec 0.3.3post6
implementing:
	general RDF 0.2.0
	model RDF 0.3.3
Error in lala:
"expected loaded resource to be a dictionary, but got type <class 'str'>: lala"
```

constantinpape (approved):

PR #275, opened 2021-11-12
generate more docs and use them

FynnBe:


constantinpape (approved):
I guess this will not change anything w.r.t. the actual functionality but just add a script for generating the weights (and add that to CI). So feel free to merge whenever you think this is ready.

PR #277, opened 2021-11-22
Stardist example model

constantinpape:
Add a stardist example model with tf weights


FynnBe (approved):
the more the merrier! 👍

PR #278, opened 2021-11-23
Update halo description

FynnBe:
A description needing additional explanation is not good enough... https://github.com/stardist/stardist/pull/171#discussion_r754915876 This is an attempt to fix that.


constantinpape:
Not sure if this helps to make it more clear. Interpreting the comment from @uschmidt83 in https://github.com/stardist/stardist/pull/171, I think that we mention "crop" here might be confusing.

Maybe we could also link to the reference implementation (https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/prediction.py#L192-L234) here?

constantinpape:
Note also https://github.com/bioimage-io/core-bioimage-io-python/issues/157.

FynnBe:
> Not sure if this helps to make it more clear. Interpreting the comment from @uschmidt83 in [stardist/stardist#171](https://github.com/stardist/stardist/pull/171), I think that we mention "crop" here might be confusing.

yes, that's why I removed 'crop' and only left one 'not cropped'...

> 
> Maybe we could also link to the reference implementation (https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/prediction.py#L192-L234) here?

always a good idea 👍

FynnBe:
how about this: https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages-halo_descr/model_spec_0_3.md#outputs:halo ?

constantinpape:
I think we can go ahead with this to merge the code changes; the description is already a bit better and we can work on improving it further. (I think the proper thing here would be to include an illustration; there is a very nice one @esgomezm has that could be used for it.)

PR #279, opened 2021-11-23
add anchors to RDF docs

FynnBe:


FynnBe:
these anchors `<a name="{name}"></a>` are not doing it. 
note for future: The goal here is to enable direct links to fields via url fragments

one possible approach is making each field a header, which has the drawback of enlarging the docs with a lot of white space.
inline ending of a small header would be neat

scripts/generate_rdf_docs.py
```diff
@@ -111,7 +111,7 @@ def markdown_from_doc(doc: DocNode, indent: int = 0):
 
             print("name", name)
 
-        name = f"`{name}` " if name else ""
+        name = f'<a name="{name}"></a>`{name}` ' if name else ""
```
oeway, line 114: Maybe you can try `id=`, see here: https://stackoverflow.com/a/484781

FynnBe: thanks @oeway let's give that a try...

FynnBe: works like a charm!

FynnBe:
I also added "unique" subfield anchors, e.g.
https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages-field_links/model_spec_latest.md#inputs:axes
https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages-field_links/model_spec_latest.md#outputs:axes

-> Every subfield has a "path" with colon separated field names.

PR #281, opened 2021-11-23
Fix shape in stardist rdf and add rdf_wrong_shape

constantinpape:


FynnBe (approved):
awesome 👍 one more test case!

PR #282, opened 2021-11-23
Float offset

### LINKED ISSUE #280: offset in ImplicitOutputShape is cast to int
This is an issue because we calculate the output shape as `out_shape = in_shape * scale + offset * 2`, so we need to allow for fractional output values in case the target value for `offset * 2` is an odd number.

FynnBe:
closes #280

enforcing that offset is a multiple of 0.5

### DIFF
--- a/bioimageio/spec/shared/raw_nodes.py
+++ b/bioimageio/spec/shared/raw_nodes.py
@@ -127,8 +127,8 @@ class Dependencies(RawNode):
 
 @dataclass
 class ParametrizedInputShape(RawNode):
-    min: List[float] = missing
-    step: List[float] = missing
+    min: List[int] = missing
+    step: List[int] = missing
 
     def __len__(self):
         return len(self.min)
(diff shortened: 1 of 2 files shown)


constantinpape:
The changes look good, but the tests now all fail. Do we maybe enforce `float` values with these changes?
That should not be the case! Maybe we need `List[Union[float, int]]`?

FynnBe:
ints are perfectly valid floats so we don't need any Union there. The whole list of floats threw the initial validation off...
anyhow, this new error `ModuleNotFoundError: No module named 'conda._vendor.auxlib'` is very obscure and seems unrelated!?

constantinpape:
> ints are perfectly valid floats so we don't need any Union there. The whole list of floats threw the initial validation off...
> anyhow, this new error `ModuleNotFoundError: No module named 'conda._vendor.auxlib'` is very obscure and seems unrelated!?

I restarted the tests. Let's see if it happens again, but yes it looks unrelated.

constantinpape:
Ok, it's still failing.
@k-dominik any idea what might cause the test failure in the conda-build test:
```
  File "/usr/share/miniconda/lib/python3.9/site-packages/conda_build/environ.py", line 483, in meta_vars
    d.update(get_git_info(git_exe, git_dir, meta.config.debug))
  File "/usr/share/miniconda/lib/python3.9/site-packages/conda_build/environ.py", line 191, in get_git_info
    from conda._vendor.auxlib.packaging import _get_version_from_git_tag
ModuleNotFoundError: No module named 'conda._vendor.auxlib'
```

PR #284, opened 2021-11-23
Add halo to stardist example model

constantinpape:
to use it in tests for tiling.

This is a very minor change, so I will just merge it if all tests (except for conda build which is expected to fail) pass.


constantinpape:
Weird, now all test pass ....

PR #285, opened 2021-11-23
Update VERSION

constantinpape:
To enable models with float offsets (needed for stardist)


constantinpape:
@FynnBe any objections against making this release?

FynnBe:
no, go ahead. we could merge https://github.com/bioimage-io/spec-bioimage-io/pull/279 before (fails only due to the conda-build issue), but it's only docs related anyhow

PR #288, opened 2021-11-25
don't use URI as path

FynnBe:
motivated by persistent issues around the use of raw_nodes.URI as path: https://github.com/bioimage-io/core-bioimage-io-python/pull/163#issuecomment-979185870


constantinpape (commented):
These are the parts of the spec that I don't really understand, so I can't review much here. Feel free to merge this when you think the issues that we are seeing are fixed.

Issue #296, opened 2021-11-29
Finalizing `v0.4`


constantinpape:
- remove `language` and `framework` fields (not useful)
- move `source` and `kwargs` to pytorch weight format
- require that tensor names are unique across input AND output tensors

constantinpape:
- for tensorflow models the input output tensor names must agree with the names of the input/output tensor in the graph in order to be compatible with Java (this was true beforehand already, but not fully documented.)

FynnBe:
last time we discussed the unique tensor names we also intended adding:
 - model inputs (like outputs) may be defined by `scale`ing and `offset`ing a `reference_tensor` (#236)

FynnBe:
@constantinpape could we finalize the halo description (#278) as well?

constantinpape:
> last time we discussed the unique tensor names we also intended adding:
> 
>     * model inputs (like outputs) may be defined by `scale`ing and `offset`ing a `reference_tensor` (

Yes, good point, we should include that.

constantinpape:
@FynnBe @oeway  I forgot something else we discussed on friday: being strict about the SPDX licences
I think we should discuss this with the others first though, so I would not make this change just yet.

FynnBe:
right, let's bring it up tomorrow morning 👍 I'll prepare it as a separate PR

oeway:
Ok, added SPDX to the spreadsheet

FynnBe:
in the weights entry for `pytorch_state_dict` (as for any weights entry) the field `source` already exists (pointing at the source of the weight file. 
I therefore propose to call the the `source` field we want to move in there `code` instead to avoid the name collision.

FynnBe:
or should we go with `source_code`?

constantinpape:
done

PR #297, opened 2021-11-29
Implementations for v0.4

FynnBe:
Implements changes summarized in https://github.com/bioimage-io/spec-bioimage-io/issues/296 and some additional minor changes

- [x] remove language and framework fields (not useful)
- [x] move source and kwargs to pytorch weight format
- [x] require that tensor names are unique across input AND output tensors
- [x] only accept SPDX license identifiers

additional minor changes: 
- add warnings:
   - [ ] length of resource name (above 64) (separate PR...)
   - [ ] SPDX free license and OSI approved? (separate PR...)
- [x] update description of parent:sha256 
- [x] allow `icon` field (see #298)


FynnBe:
we have the `icon` field in the general RDF, thus we should allow it for the model RDF to make it a valid general RDF.

constantinpape (commented):
Looks good in principle. But I think we should update a few more of the example models (e.g. multiple outputs and fixed output shape unets to v0.4).

PR #300, opened 2021-11-30
WIP update examples

FynnBe:


constantinpape:
@FynnBe  I have updated all the examples now, but for 2 models (`unet2d_diff_output_shape` and `unet2d_multi_tensor`) loading fails with:
```
   bioimageio/spec/io_.py:221: in load_raw_resource_description
      raw_rd = schema.load(data)
  .tox/py37/lib/python3.7/site-packages/marshmallow/schema.py:720: in load
      data, many=many, partial=partial, unknown=unknown, postprocess=True
  .tox/py37/lib/python3.7/site-packages/marshmallow/schema.py:865: in _do_load
      error_store=error_store, data=result, many=many
  .tox/py37/lib/python3.7/site-packages/marshmallow/schema.py:1142: in _invoke_field_validators
      error_store=error_store,
  .tox/py37/lib/python3.7/site-packages/marshmallow/schema.py:493: in _call_and_store
      value = getter_func(data)
  bioimageio/spec/model/v0_4/schema.py:316: in no_duplicate_output_tensor_names
      names = [t.name for t in value]
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  .0 = <list_iterator object at 0x7f48f7d2a590>
  
  >   names = [t.name for t in value]
  E   AttributeError: 'dict' object has no attribute 'name'
  ```
  I am not sure what's going on there.

FynnBe:
yes, this is actually expected according to the invocation order of decorated marshmallow schema methods: https://marshmallow.readthedocs.io/en/stable/extending.html
(the creation of the raw_nodes is a @post_load step)

However, that this worked so far for other models is a bug!?!?!

FynnBe:
the problem is, when I fix it and treat the data to be validated as a list of dicts (the raw data from the yaml), then I get:
`TypeError: 'InputTensor' object is not subscriptable`
-> the order of @validates and @post_load decorated methods is not always the same! 👎

FynnBe:
The issue was harder to debug due to the use of a schema class in a nested field. To make the code easier to understand I decided to enforce the use of fields and schema instances throughout the module.

FynnBe:
The real reason for the error is a validation error in the output, which causes the validation to occur on the output data and not the output node (as its creation failed). The fix is to expect both a dict and the expected raw_node, knowing that there was an upstream error if a dict "arrives"...

constantinpape:
Now it says `marshmallow.exceptions.ValidationError: {'inputs': ['Could not check for duplicate input tensor names due to another validation error.'], 'outputs': ['Could not check for duplicate output tensor names due to another validation error.']}`. Which is a bit more helpful, but it would be even better to tell us which other error occured ;).

PR #304, opened 2021-11-30
Remove Fiji from the consumer

oeway:
@FynnBe We decided to remove fiji for now from the model card and downloader


FynnBe:
the line deleted can go as well, but the deciding part for 'fiji' to disappear is specified by 
`MANIFEST_URL = "https://raw.githubusercontent.com/bioimage-io/bioimage-io-models/gh-pages/manifest.bioimage.io.json"`
So we might have to update something there...

PR #305, opened 2021-11-30
Add missing import

constantinpape:


bioimageio/spec/model/raw_nodes.py
```diff
@@ -1,3 +1,6 @@
 # Auto-generated by generate_passthrough_modules.py - do not modify
 
 from .v0_4.raw_nodes import *
+
+# * imports don't include classes prefixed with _ so we need to explicitly import the * classes we need here
+from .v0_4.raw_nodes import _WeightsEntryBase
```
FynnBe, line 6: Can't we fix this by not using this private base class? in an `isinstance` check we can do `get_args(WEightsEntry)` instead (`WeightsEntry` is a Union of the actual possible weights entries)

PR #306, opened 2021-11-30
Minor validation fixes

FynnBe:


bioimageio/spec/model/v0_4/schema.py
```diff
@@ -424,24 +437,53 @@ def get_tnames(tname: str):
...
 
         valid_input_tensor_references = get_tnames("inputs")
-        # a model is not allowed to have no or empty outputs, but I hope that this is validated somewhere else,
-        # so I use get with a default empty list here
-        for out in data.get("outputs", []):
-            if out.postprocessing is missing_:
+        ins = data.get("inputs", [])
+        outs = data.get("outputs", [])
+        if not isinstance(ins, list) or not isinstance(outs, list):
+            raise ValidationError(
```
constantinpape, line 443: Should this ever happen? I.e. shouldn't the other ValidationErrors be encountered earlier?
If that's the case, then we should maybe throw some other error here, prob. `RuntimeError`.

constantinpape (approved):
Looks good overall. It might make more sense to throw RuntimeErrors if the later validation schemes, but I don't know if we can rely on the validation order for this.

bioimageio/spec/model/v0_4/schema.py
FynnBe: the code is executed anyway, because we generate the dict structure with all errors and don't just raise on the first...

FynnBe:
> Looks good overall. It might make more sense to throw RuntimeErrors if the later validation schemes, but I don't know if we can rely on the validation order for this.

RuntimeError wouldn't work. It needs to be (or inherite from) marhsmallow.ValidationError

bioimageio/spec/model/v0_4/schema.py
constantinpape: ok, then it should be fine..

PR #307, opened 2021-11-30
Move sha256 field for pytorch weights

FynnBe:
also enforces the field (now called `architecture_sha256` in the pytorch model dict weights if `architecture` points to a file (and not a module).

todo:
 - [ ] fix: Enforcing it invalidates some of the older models we still use in testing..


FynnBe:
remaining issues should be fixed ones the updated examples are on main (there are urls in tests pointing to them on main)
for updating core we need to merge this

FynnBe:
only url tests pointing to main fail. sould pass after merge...

PR #311, opened 2021-12-01
Add update-format command

constantinpape:


constantinpape:
calling it with `str` instead does not work, because then it fails internally with
```
update_format failed with 'str' object has no attribute 'suffix'
```

FynnBe:
> Not sure why, looks like Path does not count as os.PathLike?

no, the `Union[str,.. ` part is the problem

FynnBe:
calling `save_raw_resource_description` with a `Path` should fix it...

bioimageio/spec/commands.py
```diff
@@ -5,12 +5,20 @@
...
 
 KNOWN_COLLECTION_CATEGORIES = ("application", "collection", "dataset", "model", "notebook")
 
 
+def update_format(rdf_source: Union[dict, os.PathLike, IO, str, bytes], path: Union[os.PathLike, str]):
+    """Update a BioImageie.IO"""
+    raw = load_raw_resource_description(rdf_source, update_to_current_format=True)
+    if raw.type != "model":
+        raise NotImplementedError(f"update_format is currently only for supported for models, not {raw.type}")
+    save_raw_resource_description(raw, path)
```
FynnBe, line 19:
```suggestion
    save_raw_resource_description(raw, Path(path))
```

constantinpape:
All tests pass now, so I will merge.

PR #313, opened 2021-12-01
flat warnings

FynnBe:
flat warnings work fine. they are just flat in the sense that they only contain the field name and no information about how this field is nested within the spec.


constantinpape (approved):
Changes look good, feel free to merge.

PR #314, opened 2021-12-01
rename weights format pytorch_script to torchscript

FynnBe:
and fix unet2d_braod_nuclei version conversion tests (were broken by new maintainers field)

### DIFF
--- a/bioimageio/spec/model/v0_4/raw_nodes.py
+++ b/bioimageio/spec/model/v0_4/raw_nodes.py
@@ -16,11 +16,9 @@ from bioimageio.spec.model.v0_3.raw_nodes import (
     PreprocessingName,
     Postprocessing,
     Preprocessing,
-    PytorchScriptWeightsEntry,
     RunMode,
     TensorflowJsWeightsEntry,
     TensorflowSavedModelBundleWeightsEntry,
-    WeightsFormat,
     _WeightsEntryBase,
 )
 from bioimageio.spec.rdf.v0_2.raw_nodes import CiteEntry, Dependencies, RDF as _RDF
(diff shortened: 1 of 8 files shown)


bioimageio/spec/model/v0_4/raw_nodes.py
```diff
@@ -48,6 +46,9 @@
 PreprocessingName = PreprocessingName
 
 FormatVersion = Literal["0.4.0"]  # newest format needs to be last (used in __init__.py)
+WeightsFormat = Literal[
+    "pytorch_state_dict", "torchscript", "keras_hdf5", "tensorflow_js", "tensorflow_saved_model_bundle", "onnx"
```
constantinpape, line 50: Does the order here matter? Does it determine the weight preferences in core?

scripts/generate_weight_formats_overview.py
```diff
@@ -14,9 +14,9 @@
 
 # defaults for transition period
 consumer_defaults = {
-    "ilastik": ["pytorch_state_dict", "onnx", "pytorch_script"],
+    "ilastik": ["pytorch_state_dict", "onnx", "torchscript"],
     "zero": ["keras_hdf5"],
-    "deepimagej": ["pytorch_script", "tensorflow_saved_model_bundle"],
+    "deepimagej": ["tensorflow_saved_model_bundle"],
```
constantinpape, line 19: We shouldn't drop `torchscript` from deepimageJ, we need to support it there in order to demonstrate compatibility with ilastik.
(They may need to update for this one though, but I think that's ok as it shouldn't concern most of the older models.)

constantinpape (commented):
Looks good overall, but we shouldn't remove torchscript from deepIJ

constantinpape:
Test failures are only against models on main, so this shouldn't matter.

bioimageio/spec/model/v0_4/raw_nodes.py
FynnBe: no, core has its own list (as it also may have a shorter list). It needs a separate update here: https://github.com/bioimage-io/core-bioimage-io-python/blob/1200b5a9d9b51430217c1adea88929f880b4e038/bioimageio/core/prediction_pipeline/_model_adapters/_model_adapter.py#L10

constantinpape:
All relevant tests are passing, I am gonna merge it.

Issue #265, opened 2021-10-20
package extra `full`


FynnBe:
We had a discussion about adding where/how to add the `ruamel.yaml` dependency, which is already included in the `test` extra.  @oeway suggested a new extra: `full`

`test` also includes pytest, tox and mypy. so maybe adding `full` is justified...

_Originally posted by @FynnBe in https://github.com/bioimage-io/spec-bioimage-io/issues/259#issuecomment-946493022_

oeway:
Hi @FynnBe I recently did some fix for another package which meant for both native and pyodide, I think the best solution so far is to use the conditional statement.

```python
setup(
    ...,
    install_requires=[
        ...
        "ruamel.yaml;platform_system!='Emscripten'", # will be included for native python
    ],
)
```
See here: https://github.com/pyodide/pyodide/issues/1984#issuecomment-974888357

k-dominik:
naive question @oeway: is there a way to test the package with pyodide in CI?

oeway:
@k-dominik Sure, basically we can install a browser in CI and use that to run a webpage to test pyodide. I was doing this with `playwright` which also has a python version. This does mean it requires some efforts to set it up.

constantinpape:
Issue reported by @esgomezm:
The validation command in the doc: https://github.com/bioimage-io/spec-bioimage-io#bioimageio-spec-validator does not work as expected.
This is because yaml is not included in the default installation (see discussions in this issue).
We should also somewhere add a recommendation to use `bioimageio.core` if one has access to  a full python installation, as this avoids many issues.

oeway:
I believe we already have a solution for this using the conditional
requirement in the setup.py file. @Fynn

FynnBe:
closing as ruyaml has been a required spec dependency for quite a while now..

PR #316, opened 2021-12-02
update dependencies

FynnBe:
always include ruamel.yaml when on native python (not pyodide) as discussed in #265

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
         "typing-extensions",
     ],
     entry_points={"console_scripts": ["bioimageio = bioimageio.spec.__main__:app"]},
-    extras_require={"test": ["pytest", "tox", "mypy"], "dev": ["pre-commit"]},
+    extras_require={"test": ["pytest", "tox", "mypy", "numpy"], "dev": ["pre-commit"]},
     scripts=[
         "scripts/generate_json_specs.py",
         "scripts/generate_processing_docs.py",
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup(
     ],
     packages=find_namespace_packages(exclude=["tests"]),  # Required
     install_requires=[
+        "numpy",
         "marshmallow-jsonschema",
         "marshmallow-union",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,17 +26,16 @@
...
         "marshmallow-union",
+        "marshmallow>=3.6.0,<4.0",
         "requests",
+        "ruamel.yaml;platform_system!='Emscripten'",
         "typer",
         "typing-extensions",
     ],
     entry_points={"console_scripts": ["bioimageio = bioimageio.spec.__main__:app"]},
-    extras_require={"test": ["pytest", "tox", "torch", "numpy", "mypy", "ruamel.yaml"], "dev": ["pre-commit"]},
+    extras_require={"test": ["pytest", "tox", "mypy"], "dev": ["pre-commit"]},
```
constantinpape, line 38: It seems like we can't remove numpy here, se test failutes.

FynnBe: right, I just blindly gave it a try... pytorch should not be required anymore though...

FynnBe:
conda is not yet on board: `Error message: Invalid version ''emscripten'': invalid character(s)`

FynnBe:
a bit of brute force here: ebbd252 
@oeway when using pyodide you only use pip, right?
I suppose there would be a way to get a proper selector for conda meta.yaml, but if we don't have a use case for it (yet), why make things complicated... ?!^^

setup.py
```diff
@@ -26,17 +26,17 @@
...
     packages=find_namespace_packages(exclude=["tests"]),  # Required
     install_requires=[
-        "PyYAML>=5.2",
-        "imageio>=2.5",
-        "marshmallow>=3.6.0,<4.0",
+        "numpy",
         "marshmallow-jsonschema",
         "marshmallow-union",
+        "marshmallow>=3.6.0,<4.0",
         "requests",
```
oeway, line 33: This one shouldn't be in the pyodide
```suggestion
        "requests;platform_system!='Emscripten'",
```

PR #318, opened 2021-12-02
Warnings update

FynnBe:
- better validation error on attempt of deserializing a sub schema with anything other than a dict.
- warn about missing tensorflow_version and opset_version
- don't warn about OSI approval


constantinpape (approved):
Looks good, feel free to merge when you think this is ready.

PR #325, opened 2021-12-06
Update general documentation field to allow URLs

FynnBe:
for model RDF we keep accepting only local relative markdown files (which are also allowed for the general RDF)

With this change we usually would increment the general RDF patch format version, but this should more be seen as a bug fix of the current spec as it simply did not reflect our use of the general spec so far.


oeway:
Looks good to me! Yes, we can see this as a bug fix.

constantinpape (approved):

PR #327, opened 2021-12-06
Clean up

FynnBe:
various minor clean up commits


constantinpape (commented):
Changes look good, but could we also address https://github.com/bioimage-io/spec-bioimage-io/issues/329 here?

PR #328, opened 2021-12-06
add maintainers to RDF

FynnBe:
on top of #327

PR #330, opened 2021-12-06
Separate collection

FynnBe:
on top of #328

to properly document and validate the collection RDF this PR moves the specification of the collection RDF into its own submodule. The format version stays synced with the general RDF.


FynnBe:
As we are not expecting any models in any current collections at all this PR stops the special treatment of `model` entries in the collection that consistet only of a `download_url`.

constantinpape (commented):
I think moving collections to its own submodule makes sense. I haven't looked at the changes in much details as they are intertwined with the previous PRs, but feel free to merge when you think this is ready to go.

PR #332, opened 2021-12-06
Dump unknown for attachments and collection entries

FynnBe:
Currently loading attachments and collection entries will keep unknown fields in the respective node's `unknown` attribute, but on dumping that data is lost. This PR fixes this.


constantinpape (approved):
Looks good

PR #335, opened 2021-12-07
Add update_to_format kwarg and improve changelog

FynnBe:


bioimageio/spec/io_.py
```diff
@@ -206,6 +206,7 @@ def _get_spec_submodule(type_: str, data_version: str = LATEST) -> SpecSubmodule
 def load_raw_resource_description(
     source: Union[dict, os.PathLike, IO, str, bytes, raw_nodes.URI, RawResourceDescription],
     update_to_current_format: bool = False,
+    update_to_format: Optional[str] = None,
```
constantinpape, line 209: I find two parameters here a bit confusing, especially since we can have contradicting values, e.g. `update_to_current_format=True, update_to_format=0.3`.

constantinpape: Maybe we could have a single parameter `update_to_format: Union[bool, str] = False`, which updates to the latest format if set to true.

constantinpape (commented):
There should also be tests for updating to format versions that are not latest.

bioimageio/spec/io_.py
FynnBe: yes `update_to_format` is totally sufficient, I didn't want to change the api though. maybe we should deprecate `update_to_current_format`?

constantinpape: I think we should just change the API, given that there is now big outside adoption of the library yet and I doubt that anyone would rely on a flag with default value here. (Except maybe @k-dominik, but he can manage ;).)

constantinpape: (If we don't change it now it will stick around and confuse potential users)

FynnBe: very good. I'll remove it

FynnBe:
> There should also be tests for updating to format versions that are not latest.

there is now a test updating to 0.3 and 0.4 for all unet2d models that are less or equal. other combinations don't exist yet as only the latest patch versions are available to update to (which actually happens even if specifying not to update at all)

FynnBe:
small addendum here: https://github.com/bioimage-io/spec-bioimage-io/pull/337

PR #343, opened 2021-12-11
Better io

FynnBe:
move some io functionality from core here to reduce duplicate code and make spec slightly more useful.

Biggest change is that `load_raw_resource_description` now behaves like it does in core, i.e. returns urls for fields containing local relative files if the source is a url (otherwise relative paths are returned, which are relative to `root_path`)

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -57,9 +57,9 @@ def validate(
             source_name = str(rdf_source)
         except Exception as e:
             source_name = str(e)
-
-    if not isinstance(rdf_source, dict):
-        error = f"expected loaded resource to be a dictionary, but got type {type(rdf_source)}: {str(rdf_source)}"
+    else:
+        if not isinstance(rdf_source, dict):
+            error = f"expected loaded resource to be a dictionary, but got type {type(rdf_source)}: {str(rdf_source)}"
 
     raw_rd = None
     if error is None:
--- a/bioimageio/spec/io_.py
+++ b/bioimageio/spec/io_.py
@@ -4,18 +4,16 @@ which is a python dataclass
 """
 import os
 import pathlib
-import re
 import warnings
 import zipfile
-from io import BytesIO, StringIO
+from hashlib import sha256
+from io import StringIO
 from types import ModuleType
 from typing import Dict, IO, Optional, Sequence, Tuple, Union
 
-from marshmallow import ValidationError
-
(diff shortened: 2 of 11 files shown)


bioimageio/spec/commands.py
```diff
@@ -57,9 +57,9 @@ def validate(
             source_name = str(rdf_source)
         except Exception as e:
             source_name = str(e)
-
-    if not isinstance(rdf_source, dict):
-        error = f"expected loaded resource to be a dictionary, but got type {type(rdf_source)}: {str(rdf_source)}"
+    else:
```
constantinpape, line 60: I think `elif not ....` would be more readable here.

constantinpape (commented):
I haven't looked at the changes in much detail; seems like this is mostly moving code around.
But the tests fail with:
```
   E   ModuleNotFoundError: No module named 'tqdm'
```
So it seems that we need to add `tqdm` to the dependencies to make this work.
Not sure if this is easy for installation with micropip. cc @oeway

oeway:
Can you add a platform condition so we don't install that for pyodide? The thing is that pyodide download every time all the new libraries, so it's better to keep very few dependencies. 

@constantinpape you can find examples in the setup.py if we are already using it.

bioimageio/spec/commands.py
FynnBe: unfortunately the `else` of a `try/except` block can not be an `elif`.

FynnBe:
as `tqdm` is only used in conjunction with `requests` when downloading and I don't see it listed in https://github.com/pyodide/pyodide/tree/main/packages I added it as dep for native python only.

bioimageio/spec/commands.py
constantinpape: Maybe I am missing something, but as far as I can see this `else` is not linked to any `try/except` block at all, but only relates ti the `if not isinstance ...` above.
Afaik there is no `try/except/if/else` construction ...

constantinpape (approved):
Looks good overall.

bioimageio/spec/commands.py
FynnBe: there are two try/except blocks, the outer one has the else: https://github.com/bioimage-io/spec-bioimage-io/blob/29446bd9b8015513705445c11a507f79f6c2dbfd/bioimageio/spec/commands.py#L51-L62

PR #365, opened 2022-01-07
Simplify collection RDF

FynnBe:


constantinpape:
From the discussion in the meeting:
- update the first entry so that source links to an rdf on github, so that this is the actual use case
- id should be mandatory for each entry in collection and the ids should be unique

FynnBe:
> * update the first entry so that source links to an rdf on github, so that this is the actual use case

@oeway and I came to the conclusion that a dedicated `rdf_source` field fits our needs better....

oeway:
Looks good to me

PR #375, opened 2022-01-15
Make some fields optional in the general RDF

oeway:
For general RDF and collection, forcing authors, cite, documentation, and tags does not make sense all the time. It's much better to make these non-critical fields optional.  

For example, when the RDF item is not published anywhere, there won't be much to fill for the `authors` and `cite` fields. `tags` is helpful for searching but it's not critical, similar case for `documentation` (note we have `description` already).

Another use case is to generate and use RDFs offline, e.g. with the bioimageio.core package if we generate a model RDF for offline use, we will have to generate empty fields for `authors`, `cite`, `tags` and `documentation` which seems unnecessary. They only make sense if we want to upload the model to the model zoo, but we can always enforce them when uploading to the website or as part of our approval criterion.

The fields change is non-breaking and it means a minimal RDF file should have the following fields:
 - format_version
 - name
 - type
 - description

This PR also includes moving the general RDF examples to a separate file to make it easier to find the link for model/collection spec.


constantinpape (approved):
The changes look good to me. Do you need to have this immediately or can we wait with merging until @FynnBe had a chance to look at this, who is more familiar with this part of the code?

oeway:
All the tests passes now, I will merge this now to get all the collections working on the website. If @FynnBe spotted anything later we can patch it with another PR.

PR #384, opened 2022-01-26
Add the validator plugin for the uploader

oeway:
This PR brings the validator plugin that runs on the bioimage.io uploader here, so we can easily make the versions sync.

FYI: @FynnBe @constantinpape Next time when we bump the version, we need to also change this file to upgrade the bioimageio.spec package used for the uploader.

### DIFF
--- a/scripts/bio-rdf-validator.imjoy.html
+++ b/scripts/bio-rdf-validator.imjoy.html
@@ -1,0 +1,43 @@
+<config lang="json">
+{
+  "name": "BIO-RDF-Validator",
+  "type": "web-python",
+  "version": "0.1.0",
+  "description": "A spec validator for bioimage RDF files.",
+  "tags": [],
+  "ui": "",
+  "cover": "",
+  "inputs": null,
+  "outputs": null,
+  "flags": [],
+  "icon": "extension",
+  "api_version": "0.1.8",
+  "env": "",
+  "permissions": [],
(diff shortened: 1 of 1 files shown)


scripts/bio-rdf-validator.imjoy.html
```diff
@@ -0,0 +1,43 @@
...
+  "ui": "",
+  "cover": "",
+  "inputs": null,
+  "outputs": null,
+  "flags": [],
+  "icon": "extension",
+  "api_version": "0.1.8",
+  "env": "",
+  "permissions": [],
+  "requirements": ["bioimageio.spec==0.4.3"],
```
oeway, line 17: We need to change here when we have a new version for the uploader.

constantinpape:
It's good to have this one here.

scripts/bio-rdf-validator.imjoy.html
constantinpape: Can we somehow read this from https://github.com/bioimage-io/spec-bioimage-io/blob/main/bioimageio/spec/VERSION instead?

oeway: How can that be different with not pin the version? I think the point to pin is that we can bump it manually and not break the website even when the latest version does work as expected.

constantinpape: ah ok, I see. Yes then let's keep the pin here and bump it manually when needed.

PR #388, opened 2022-02-01
more informative validation summary name

FynnBe:


bioimageio/spec/commands.py
```diff
@@ -77,6 +79,10 @@ def validate(
                 error = str(e)
                 tb = traceback.format_tb(e.__traceback__)
 
+            if raw_rd is not None:
+                format_version = raw_rd.format_version
+                resource_type = "general" if raw_rd.type == "rdf" else raw_rd.type
```
oeway, line 84: Just curious, in what case raw_rd.type will have type "rdf"?

FynnBe: in the general case..  "general" is not official anywhere, 'rdf' is the default type. We might want to change that, but it doesn't matter all that much

constantinpape (approved):

Issue #392, opened 2022-02-15
validate minimum shape produces a "valid" output after processing


k-dominik:
Parametrized input shape is currently defined as: `shape = min + k * step for k in {0, 1, ...}`
The output shape is defined as `shape = shape(input_tensor) * scale + 2 * offset` as well as `shape_after_crop = shape - 2 * halo`. 
As a tiling consumer software, one would _always_ remove halo. So, it is possible, that for some `k` the output shape could be `<=0` (for some dimensions), which I would say constitutes an invalid model. Could we enforce that the input shape has to be given such that the output after halo cropping is still a valid shape?

constantinpape:
I agree that this should be checked. It might not be possible to check this statically though (which is implemented here), but rather only possible in the dynamic check (in https://github.com/bioimage-io/core-bioimage-io-python). But @FynnBe would know how to best do this.

FynnBe:
we should be able to do this statically....

PR #395, opened 2022-02-18
Improve output shape validation

FynnBe:
as discussed in #392


constantinpape (commented):
We should add a breaking example and add a test that uses it to make sure that the issue is caught.

constantinpape (approved):
Sorry, nvmd I just saw that it's there already. (You just produce the example on the fly instead of storing it; and that's fine.)

k-dominik (approved):
looks great! :)

PR #396, opened 2022-02-21
Warn about tags

### LINKED ISSUE #315: Warn if there is not a tag for each category
https://github.com/bioimage-io/bioimage.io/blob/main/site.config.json#L109-L140

FynnBe:
fixes #315

if possible check the website config for tag categories and warn if tag categories are missing in the tags specified.

Unfortunately there were some bad (near cyclic) imports in bioimageio.spec.shared.utils from bioimageio.spec.shared. This PR also cleans that up, which breaks a few imports in core (some objects have moved from bioimageio.spec.shared.utils to bioimageio.spec.shared\[.node_transformer\] ). fix in core: https://github.com/bioimage-io/core-bioimage-io-python/pull/227

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ bioimageio update-format <MY-MODEL-SOURCE> <OUTPUT-PATH>
 
 
 ## Changelog
+#### todo: bioimageio.spec 0.4.3post1
+  - changes to api:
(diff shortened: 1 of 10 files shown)


constantinpape (approved):
The warnings look good to me. I haven't checked the changes in imports but I guess this should be fine since we only need to update in `bioimageio.core`.

PR #398, opened 2022-02-21
Docs on demand

### LINKED ISSUE #372: Don't build doc strings by default.
There are some issues caused by the fact that the docstring are build whenever a resource is loaded, see for example the issues here: https://github.com/bioimage-io/bioimage.io/issues/216.
It would be much cleaner if this can be separated from running validation and they are only built on demand, e.g. when the doc is actually generated.

FynnBe:
fixes #372

`biomageio_descriptions` that are not simple strings are wrapped in lambda functions and only called in the doc building scripts...


constantinpape (approved):
Looks good, feel free to merge.

PR #401, opened 2022-02-22
Fix training data field

FynnBe:


bioimageio/spec/model/v0_4/schema.py
```diff
@@ -527,7 +531,7 @@ def get_min_shape(t) -> numpy.ndarray:
         "(#https://en.wikipedia.org/wiki/ISO_8601) format.",
     )
 
-    training_data = fields.Union([fields.String(bioimageio_description="dataset id"), fields.Nested(Dataset())])
+    training_data = fields.Union([fields.Nested(Dataset()), fields.Nested(LinkedDataset())])
```
FynnBe, line 534: @oeway @constantinpape here we check first for an in-place dataset rdf and otherwise for the LinkedDataset that currently only consists of `id`. unknown fields are not allowed for the `LinkedDataset`

FynnBe:
tests seem to fail only due to links to main branch where the example model is not updated yet...

constantinpape (approved):
Good to go from my side. We can just merge despite the test errors, they should sort themselves out once the file is updated on main; if not we can fix it in a follow-up PR.

PR #423, opened 2022-03-16
Add no-cache env var 

FynnBe:


bioimageio/spec/shared/common.py
```diff
@@ -33,9 +33,34 @@ def dump(self, data, stream=None, *, transform=None):
...
+                yield from self.iterable
+
+        def update(self, *args, **kwargs):
+            pass
+
+
 BIOIMAGEIO_CACHE_PATH = pathlib.Path(
     os.getenv("BIOIMAGEIO_CACHE_PATH", pathlib.Path(tempfile.gettempdir()) / "bioimageio_cache")
 )
+BIOIMAGEIO_NO_CACHE = bool(os.getenv("BIOIMAGEIO_NO_CACHE", "False").title())
```
constantinpape, line 57: I don't think that this works as expected since `bool("False")` and `bool("0")` (the two most obvious values to set this to false via env vars) both evaluate to true.
I would also vote to rename this variable to `BIOIMAGEIO_USE_CACHE`, and set its default to true. This is much easier to understand than 'no_cahce" and default false; since the default is currently a double negation.

constantinpape (changes requested):
Two changes requested:
- I don't think that reading the env var to boolean works like this
- `BIOIMAGEIO_USE_CACHE` would make the code easier to understand

bioimageio/spec/shared/common.py
FynnBe: very much on point! 👍

oeway: Thanks for implementing this @FynnBe and I agree with @constantinpape for these points.

FynnBe: yeah.. looking at it today I don't know what I was thinking yesterday^^

bioimageio/spec/shared/common.py
```diff
@@ -33,9 +33,34 @@ def dump(self, data, stream=None, *, transform=None):
...
+                yield from self.iterable
+
+        def update(self, *args, **kwargs):
+            pass
+
+
 BIOIMAGEIO_CACHE_PATH = pathlib.Path(
     os.getenv("BIOIMAGEIO_CACHE_PATH", pathlib.Path(tempfile.gettempdir()) / "bioimageio_cache")
 )
+BIOIMAGEIO_USE_CACHE = os.getenv("BIOIMAGEIO_USE_CACHE", "true").lower() in ("false", "no")
```
constantinpape, line 57:
```suggestion
BIOIMAGEIO_USE_CACHE = os.getenv("BIOIMAGEIO_USE_CACHE", "true").lower() not in ("false", "no", "0")
```

constantinpape: You need the "not" here so that this evaluates correctly: to make this clear the code **without** not would read like this for the default value
```python
use_cache = "true" in ("false", "no")
# use cache evaluates to False in this case, which is wrong!
```

And I would add 0 here since it's very common to specify env variables by integers.

constantinpape (commented):
Looks better now, but the env var parsing is still wrong ;).

bioimageio/spec/shared/common.py
FynnBe: good catch! but then let's get rid of the double negative here, too 🚀 
```suggestion
BIOIMAGEIO_USE_CACHE = os.getenv("BIOIMAGEIO_USE_CACHE", "true").lower() in ("true", "yes", "1")
```

FynnBe: or do we want `BIOIMAGEIO_USE_CACHE='ja'` also to work?

constantinpape: You're suggestion looks good. Adding more synonyms is fine with me.

constantinpape (approved):
LGTM once you have fixed the env var parsing

constantinpape:
Should we merge this now @FynnBe?

PR #429, opened 2022-04-01
add update_nested

FynnBe:
as discussed in https://github.com/bioimage-io/core-bioimage-io-python/pull/256

### DIFF
--- a/bioimageio/spec/commands.py
+++ b/bioimageio/spec/commands.py
@@ -1,15 +1,28 @@
 import os
 import traceback
 import warnings
+from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, IO, Union
+from typing import Any, Dict, IO, List, Optional, Union
 
 from marshmallow import ValidationError
 
 from .collection.v0_2.utils import resolve_collection_entries
-from .io_ import load_raw_resource_description, resolve_rdf_source, save_raw_resource_description
-from .shared.common import ValidationWarning, nested_default_dict_as_nested_dict
+from .io_ import (
+    load_raw_resource_description,
+    resolve_rdf_source,
+    save_raw_resource_description,
+    serialize_raw_resource_description_to_dict,
+)
+from .shared import update_nested
+from .shared.common import ValidationWarning, nested_default_dict_as_nested_dict, yaml
+from .shared.raw_nodes import ResourceDescription as RawResourceDescription
 from .v import __version__
+
+try:
+    from typing import TypedDict, Literal
+except ImportError:
+    from typing_extensions import TypedDict, Literal  # type: ignore
 
 
 def update_format(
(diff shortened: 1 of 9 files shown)


constantinpape (commented):
The changes look good, but this is missing some user-facing documentation (and functionality?):
How can we use this to directly update a rdf.yaml file? 
Do we need additional functionality for this? If yes, should it go into core?

FynnBe:
Yes, let's maybe add a cli command `update-rdf` with the (default?) option `--validate` that takes RDF, UPDATE, OUTPUT as arguments. 
The other use cases I see for now are:
 - updates in collection (internal)
 - updates in `export_bioimageio_model` (internal)

constantinpape:
> Yes, let's maybe add a cli command `update-rdf` with the (default?) option `--validate` that takes RDF, UPDATE, OUTPUT as arguments.

That sounds good, but `update_rdf` should also be available as a proper function in `bioimageio.spec`, so that it can be used programatically and not only through the CLI.

FynnBe:
> > Yes, let's maybe add a cli command `update-rdf` with the (default?) option `--validate` that takes RDF, UPDATE, OUTPUT as arguments.
> 
> That sounds good, but `update_rdf` should also be available as a proper function in `bioimageio.spec`, so that it can be used programatically and not only through the CLI.

how about
```python
def update_rdf(
    source: Union[dict, os.PathLike, IO, str, bytes],
    update: Union[dict, os.PathLike, IO, str, bytes],
    output: Union[dict, os.PathLike] = None,
    validate: bool = True,
) -> Union[dict, Path]:
   ...
```
the default would be to return a dict

constantinpape:
> how about
> 
> ```python
> def update_rdf(
>     source: Union[dict, os.PathLike, IO, str, bytes],
>     update: Union[dict, os.PathLike, IO, str, bytes],
>     output: Union[dict, os.PathLike] = None,
>     validate: bool = True,
> ) -> Union[dict, Path]:
>    ...
> ```
> 
> the default would be to return a dict

Yeah, that's good!
Can you implement s.t. "inplace" updates are possible? I.e. if `ouptut=source` the input rdf file should be updated directly.

bioimageio/spec/commands.py
```diff
@@ -120,3 +160,67 @@ def validate(
...
+        output:  dict or path to write output to (default: return new dict)
+        validate_output: whether or not to validate the updated RDF
+
+    Returns:
+        The updated content of the source rdf as dict or,
+        if output is a path, that path (where the updated content is saved to).
+
+    Raises:
+        ValidationError: if `validate_output` and the updated rdf does not pass validation
+    """
```
FynnBe, line 184: very close to the initial draft, I only added `RawResourceDescription` for source and update, so one does not have to go back to dicts if the RDF(s) are already loaded.

constantinpape (approved):
LGTM

PR #433, opened 2022-04-26
Resolve relative paths when loading a (raw) resource

FynnBe:
this PR makes sure we don't forget a resource location when manipulating a loaded resource or loading a nested resource as is the case in a collection RDF with entries that specifie `rdf_source`. 

When serializing the user has the option to keep absolute paths or to convert them to relative ones (should be used when saving to a yaml file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ bioimageio update-format <MY-MODEL-SOURCE> <OUTPUT-PATH>
 
 
 ## Changelog
-#### bioimageio.spec 0.4.6
+#### bioimageio.spec 0.5.4.post8
 - any field previously expecting a local relative path is now also accepting an absolute path
 - load_raw_resource_description returns a raw resource description which has no relative paths (any relative paths are converted to absolute paths).
 
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ bioimageio update-format <MY-MODEL-SOURCE> <OUTPUT-PATH>
 
 
 ## Changelog
+#### bioimageio.spec 0.4.6
+- any field previously expecting a local relative path is now also accepting an absolute path
+- load_raw_resource_description returns a raw resource description which has no relative paths (any relative paths are converted to absolute paths).
+
 #### bioimageio.spec 0.4.4post7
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -121,6 +121,10 @@ bioimageio update-format <MY-MODEL-SOURCE> <OUTPUT-PATH>
 
 
 ## Changelog
+#### bioimageio.spec 0.4.6
+- any field previously expecting a local relative path is now also accepting an absolute path
+- load_raw_resource_description returns a raw resource description which has no relative paths (any relative paths are converted to absolute paths).
```
constantinpape, line 126: These make sense and sound like they will make everything more usable.

constantinpape (commented):
I didn't go through the changes in detail, but the effect you describe sounds good and is a good change overall.
Something I noticed though: we're missing the changelog for the `0.4.5` releases. Could you add them?
And are there any changes necessary in `bioimageio.core` as follow up?

FynnBe:
> Something I noticed though: we're missing the changelog for the 0.4.5 releases. Could you add them?

no there is a note in the changelog for 0.4.5, but it is under [RDF Format Versions](https://github.com/bioimage-io/spec-bioimage-io#rdf-format-versions): https://github.com/bioimage-io/spec-bioimage-io#model-rdf-045

I now added a changelog entry to the library, as we still officially require a valid RDF to have local relative paths, but now we won't throw validation errors about it anymore. For someone writing an RDF manually nothing has changed.

> are there any changes necessary in bioimageio.core as follow up?

only in the tests: https://github.com/bioimage-io/core-bioimage-io-python/pull/259

constantinpape:
Ok, thanks for the clarification. This is good to go from my side. And feel free to follow up with a new release directly.

constantinpape (approved):

FynnBe:
alright, I bumped here to '0.4.5post8' (instead of of 0.4.6 as I noted initially in the changelog) as this is a pure library update, spec is untouched...

constantinpape:
> alright, I bumped here to '0.4.5post8' (instead of of 0.4.6 as I noted initially in the changelog) as this is a pure library update, spec is untouched...

Makes sense!

PR #438, opened 2022-05-03
resolve symlinks when transforming absolute to relative paths during serialization

FynnBe:
thanks @uschmidt83 for raising this issue in https://github.com/bioimage-io/core-bioimage-io-python/issues/260


constantinpape (approved):
LGTM

Issue #252, opened 2021-10-06
Different dimensions in output vs input


constantinpape:
Currently it's not clear how to specify outputs that have a different number of dimensions in the output than in the input when using implicit shapes.
E.g. a model that gets input (a) `bczyx` and outputs `bcyx`. Or (b) `bcyx` -> `bczyx`.
I would suggest that we set offset and scale to zero for (a) and scale to zero, offset to dim_size / 2 for (b).
This way we don't need to change the spec, but probably need to adapt the core lib so it can deal with these cases.

FynnBe:
sounds good to me with one remark:
"offset to dim_size / 2 for (b)."; dim_size is not known for inputs defined with min,step. Any choice is somewhat arbitrary there, too, so maybe 0 would do the trick here, too.

oeway:
Interesting question, it feels a bit unintuitive, and it won’t work it we
swapped the dimensions in the output. Can we add another key, say,
reference_dims, so we can explicitly set the reference dimension for
computing the outputs.

Even more flexible, considering we might want to compute differen dims with
respect to different input tensor, we can perhaps change the format of
reference tensor to a list with the format like: “input_tensor_name:y”. For
example if we have two inputs tensor and we want to define the outputs by
referring to both inputs, we can do something like this:

reference_dims: [ input-1:b, input-2:x, input-2:y, input-1:z]

If it’s a newly created dimension, then we set the reference dim to null.

FynnBe:
For the straight forward case of only adding a dimension a `null` value could also be added to `scale`, with `offset` setting the length of the new dimension. 
e.g. 
```
reference_tensor: input-1
scale: [1, 1, 1, null]
offset: [0, 0, 0, 7]
```
this could be equivalent to specifying 
```
reference_dims: [input-1:b, input-1:x, input-1:y, null]
scale: [1, 1, 1, null]  # here we could also allow scale to be 0, but null is more explicit for a new dimension, so I would not allow 0 here to avoid confusion
offset: [0, 0, 0, 7]
```

Furthermore I would not allow to specify both, `reference_tensor` and `reference_dims`

esgomezm:
Hi there, any news on this? @FynnBe do you think it could be implemented??

FynnBe:
yes, but as discussed offline this does not have a high priority right now. (use of fixed output shape covers some of these use cases)

FynnBe:
fixed soon by #521

PR #442, opened 2022-05-04
Expand output tensor spec

FynnBe:
Implements changes discussed in https://github.com/bioimage-io/spec-bioimage-io/issues/252


constantinpape:
@FynnBe this should work now. Please review and go ahead if you agree.

FynnBe:
looks good. I have two follow-up questions that we could also address in a following PR, just making sure these aspects will not collide:
- [ ] how do we shrink a tensor with respect to its reference tensor?
- [ ] (should really be in another PR) we discussed in the past to allow also input tensors to be specified via another input tensor... this is not yet the case, is it??

bioimageio/spec/model/v0_4/schema.py
```diff
@@ -456,23 +456,36 @@ def get_min_shape(t) -> numpy.ndarray:
...
-                tensors_by_name[out.shape.reference_tensor].shape
-            ):
-                raise ValidationError(
-                    f"Referenced tensor {out.shape.reference_tensor} "
-                    f"with {len(tensors_by_name[out.shape.reference_tensor].shape)} dimensions does not match "
-                    f"output tensor {out.name} with {len(out.shape)} dimensions."
-                )
+            if isinstance(out.shape, raw_nodes.ImplicitOutputShape):
+                ndim_ref = len(tensors_by_name[out.shape.reference_tensor].shape)
+                ndim_out = len([scale for scale in out.shape.scale if scale is not None])
```
FynnBe, line 482: here `ndim_out` would include the `None` axis, but you compare to `ndim_ref`, so you leave it out.. maybe we can rename this here or comment on it?

bioimageio/spec/shared/schema.py
```diff
@@ -113,7 +113,9 @@ def matching_lengths(self, data, **kwargs):
 class ImplicitOutputShape(SharedBioImageIOSchema):
```
FynnBe, line 113: if we are very diligent we'd add validation such that offset is not zero where scale is `None`...

FynnBe:
@constantinpape feel free to merge if you want to leave these for a future PR ;-)

constantinpape:
> how do we shrink a tensor with respect to its reference tensor?

We haven't discussed this yet. As @oeway has correctly pointed out in https://github.com/bioimage-io/spec-bioimage-io/issues/252#issuecomment-936221835 we need a bit more work to cover all cases, which also include changing the axis order.
I suggest we skip the case of removing axes in the output for now and tackle it once we work on the more general case.

> (should really be in another PR) we discussed in the past to allow also input tensors to be specified via another input tensor... this is not yet the case, is it??

I think it's not implemented yet, but this one would be easy (just need to pass over all the tensors first and collect the names, instead of only going over the input tensors). And we need to add a few checks:
- no loop referencing of tensors (`input_a` <-> `input_b`)
- no referencing of inputs to outputs (`input` -> `output`)
- no referencing of outputs to outputs (`output_a` -> `output_b`)  (though we could also allow this one)

But it should be a separate PR.

FynnBe:
> ... But it should be a separate PR.

yes, awesome! just came back here to say that the dropping of axes could just fall into the general mixed axes case, but I see you've been faster!

Let's merge once tests pass 👍

FynnBe:
Instead of [df5ca66](https://github.com/bioimage-io/spec-bioimage-io/pull/442/commits/df5ca6650d441f5162e8c5fc618d62c223efa321) we could also update the expected error message in the test, but I thought only displaying it when relevant might avoid confusion...

PR #445, opened 2022-05-05
Upgrade the validator bioengine app to 0.4.5post13

oeway:


FynnBe:
I think the post version could not be set here.. but we can try.. it should resolve to the latest post though...
the problem causing https://github.com/bioimage-io/collection-bioimage-io/issues/406 might lie in the latest bioimageio.spec

oeway:
> I think the post version could not be set here.. 

Why?

FynnBe:
not sure, maybe I'm mixing it up with something. pip and conda pinning differs slightly... 
even if we can, I'm not sure we should pin post...

constantinpape:
Just fyi, we have just published v0.4.6. So maybe we can pin to that and don't worry about any of the post versions.

PR #446, opened 2022-05-05
Add example model with output expanstion

constantinpape:


constantinpape:
Closed in favor of #442

PR #447, opened 2022-05-06
allow root_path to be URI

FynnBe:
to resolve collection entries after the collection RDF has been loaded, we need to 'remember' the root.
Resolving everything in one go would be another approach, but would require the collection entry update logic to be fixed, which we want to avoid to support enriching the collection entries with fields from imjoy plugins.


constantinpape (approved):
LGTM

PR #450, opened 2022-05-09
Release model RDF 0.4.6

FynnBe:


constantinpape (commented):
This model needs to be bumped to 0.4.6 as well: https://github.com/bioimage-io/spec-bioimage-io/blob/main/example_specs/models/unet2d_nuclei_broad/rdf_expand_output_shape.yaml#L2

Otherwise looks good to me.

PR #451, opened 2022-05-09
Disable documentation for bioengine apps

oeway:
BioEngine apps with documentation are failing because we don't allow string in the doucmentation field: https://github.com/imjoy-team/bioimage-io-models/runs/6349058588?check_suite_focus=true#step:4:22

I need to disable this conversion now.

@FynnBe do you see a way that we can generate a local file from the string and link it to the application item?


FynnBe:
> @FynnBe do you see a way that we can generate a local file from the string and link it to the application item?

yes, that should be possible. we need to provide `root_dist` and then deploy to the same folder the updated `rdf.yaml` is deployed to... I'll take a closer look this week

FynnBe:
ok, the resources affected by this will now fail if the `documentation` field is not set manually...

PR #452, opened 2022-05-10
fix enrich_partial_rdf_with_imjoy_plugin

FynnBe:
when resolving rdf_source we cannot turn it into a path `root / rdf_source` but need to try to resolve it before as it may be a URL.


FynnBe:
this should fix the errors in collection as bubbling up here: https://github.com/bioimage-io/collection-bioimage-io/runs/6367785980?check_suite_focus=true

PR #465, opened 2022-05-31
Remove link from title

oeway:
This is needed for rendering on bioimage.io/docs


constantinpape:
I think this is the only place where we link to the rendered model RDF. Could you place the link somewhere in the text below?

oeway:
Nope, there is another link below for the latest version.

PR #472, opened 2022-09-07
load future format version as latest with warning

### LINKED ISSUE #468: allow _some_ form of forward compatibility
this comes from an offline discussion with @FynnBe.

In the context of software that is distributed in _binary_ form to end users, it is possible, that some dependencies, e.g. `bioimageio.spec` move on - which in general is fine. There is, however, a problem with models generated in newer versions - they will not work with older versions of the library. Currently, in a situation where the client is e.g. `0.4.5`, but the model was generated with `0.4.6`, the client will error out when trying to load the model. Often the models don't use any feature _specific_ to this version, but were merely generated with it.

I know there's 0ver and all of that, but it would be really great to start giving _some_ guarantees about compatibility between versions. @FynnBe already pretty much nailed backwards compatibility (so far I didn't have problems with that). So doing some kind of forward compatibility, maybe for minor versions, would be pretty useful.

CC: @oeway @constantinpape @esgomezm

FynnBe:
any format version higher than the respective type's latest known is loaded as latest instead.

fixes #468

### DIFF
--- a/bioimageio/spec/io_.py
+++ b/bioimageio/spec/io_.py
@@ -159,6 +159,7 @@ def load_raw_resource_description(
 
     class_name = get_class_name_from_type(type_)
 
+    # determine submodule's format version
     original_data_version = data.get("format_version")
     if original_data_version is None:
         odv: Optional[Version] = None
@@ -187,6 +188,7 @@ def load_raw_resource_description(
             raise e  # raise original error; no second attempt with 'LATEST' format version
 
         try:
+            # load latest spec submodule
             sub_spec = _get_spec_submodule(type_, data_version=LATEST)
         except ValueError:
             raise e  # raise original error with desired data_version
@@ -195,7 +197,7 @@ def load_raw_resource_description(
             # original format version is not a future version.
             # => we should not fall back to latest format version.
             # => 'format_version' may be invalid or the issue lies with 'type_'...
-            raise
+            raise e
 
     if odv and Version(sub_spec.format_version) < odv:
         warnings.warn(
--- a/bioimageio/spec/io_.py
+++ b/bioimageio/spec/io_.py
@@ -160,6 +160,14 @@ def load_raw_resource_description(
     class_name = get_class_name_from_type(type_)
 
     original_data_version = data.get("format_version")
+    if original_data_version is None:
+        odv: Optional[Version] = None
+    else:
+        try:
+            odv = Version(original_data_version)
+        except Exception as e:
+            raise ValueError(f"Invalid format version {original_data_version}.") from e
+
     if update_to_format is None:
         data_version = original_data_version or LATEST
     elif update_to_format == LATEST:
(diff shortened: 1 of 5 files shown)


bioimageio/spec/io_.py
```diff
@@ -170,7 +172,35 @@ def load_raw_resource_description(
...
+        except Exception as e:
+            raise ValueError(f"Invalid format version {original_data_version}.") from e
+
+        if Version(sub_spec.format_version) < odv:
+            warnings.warn(
+                f"Loading future {type_} format version {original_data_version} as (latest known) "
+                f"{sub_spec.format_version}."
+            )
+            data["format_version"] = sub_spec.format_version
+            if "config" not in data:
```
k-dominik, line 194: cool, could you comment why you add `"config"` and `"config/biomimageio"` key? just for context - I guess they were required starting from some version?

tests/test_format_version_conversion.py
```diff
@@ -39,3 +43,27 @@ def test_model_format_version_conversion(unet2d_nuclei_broad_before_latest, unet
...
+def test_forward_compatible(v_diff: Tuple[int, int, int], unet2d_nuclei_broad_latest):
+    from bioimageio.spec import load_raw_resource_description
+    from bioimageio.spec.model import format_version
+
+    fv_key = "format_version"
+
+    model_data = yaml.load(unet2d_nuclei_broad_latest)
+
+    v_latest: Version = Version(format_version)
+    v_future: str = ".".join(
```
k-dominik, line 59: maybe it's good to look at this at the end of a workday without any stamina - why is minor always incremented irrespective of v_diff?

k-dominik (commented):
oh I'm so looking forward to this - I get bitten by repacked models all the time :D

tests/test_format_version_conversion.py
FynnBe: I removed it. was left from an implementation draft. good catch!

bioimageio/spec/io_.py
FynnBe: I added some comments to clarify... 
I need to optional 'config/bioimageio' dict only because I wanted to save the original format version for reference

FynnBe: we could also not do that and remove this part. I thought it might help debugging, etc...

k-dominik: ah, no, don't remove it. My first instinct was that you added it to conform some kind of implicit schema, but if it's for debugging (and if you state that in a comment) I think it's also comprehensible for someone less familiar with the code.

k-dominik: one could even do some overkill explicit naming with `debug` something in the key... or not :)

FynnBe: debug might not be the only use case... so I left it as what it is `original_format_version` and put it in the `bioimageio` part of the config to avoid key clashes across different software (as is the recommendation)

k-dominik (approved):

Issue #470, opened 2022-07-27
Make cite optional for models


oeway:
Currently the `cite` field for the model RDF is a required field and it force to have a doi, which can be hard for unpublished models.  I think we should make it as an optional field so one can skip it. Note we already made the `cite` field in the general RDF as optional.

Also see https://github.com/bioimage-io/collection-bioimage-io/pull/473 (which has one error related to this.)

constantinpape:
I agree, I think we should throw a warning if `cite` is missing.

FynnBe:
there was a bug hiding the `url` field within a `CiteEntry`
currently `doi` or `url` have to be specified. 
We can still make `cite` opitonal of course...

PR #473, opened 2022-09-12
warn missing cite for model rdf

FynnBe:
> I agree, I think we should throw a warning if `cite` is missing.

_Originally posted by @constantinpape in https://github.com/bioimage-io/spec-bioimage-io/issues/470#issuecomment-1196837305_

I forgot about the warning. this PR adds it. **Only for the model RDF**. We could alternatively also add it for all RDFs.


constantinpape (approved):

constantinpape:
> **Only for the model RDF**. We could alternatively also add it for all RDFs.

I think only for the model RDF is fine for now.

Issue #477, opened 2022-10-26
Workflow spec proposal


oeway:
To support custom pre/post-processing, but also to allow composition of models, (together with @FynnBe and @constantinpape) we propose the follow workflow spec. 

The idea is to define a list of operators in the yaml file and chain them sequentially. Each operator is a function which takes the outputs of the previous operator as inputs.

A stardist workflow would look like this: 

```yaml
name: StarDist Example Workflow
description: A workflow for running stardist
format_version: 0.2.3
type: application
application_type: workflow
steps:
  - op: zero_mean_unit_variance
  - op: model_inference
     model_id: fearless-crab
     preprocessing: false
     postprocessing: false
  - op: stardist_postprocessing
```

To allow more complex workflow composition, we optionally support naming the operator step, the outputs which can be referred in a later step. While all the outputs of the previous step will be feed into the next step by default, we can change the inputs by specifying the input using `<step-name>.outputs.<output-index>`, or, for for named outputs from another named step, we can use `<step-name>.outputs.<output-name>`.

For example, if we define `outputs: [nuclei, cells]` in `step1`, we can refer to it in the inputs of step 2 with :`inputs [step1.outputs.cells]` or `inputs [step1.outputs.1]`.

Here is a more complete example for combing two models in one workflow: 
```yaml
name: HPA Single-cell Classification Example Workflow
description: A workflow for running HPA single-cell classification
format_version: 0.2.3
type: application
application_type: workflow
steps:
  - id: data
    op: name_tensors
    outputs: [nuclei, protein]
  - id: segmentation
    op: model_inference
    inputs: [data.outputs.nuclei]
    model_id: conscientious-seashell
    preprocessing: true
    postprocessing: false
    outputs: [cells]
  - id: classification
    op: model_inference
    inputs: [data.outputs.protein, segmentation.outputs.cells]
    model_id: straightforward-crocodile
    preprocessing: true
    postprocessing: false
    outputs: [scores]
  - op: name_tensors
    inputs: [segmentation.outputs.cells, classification.outputs.scores]
    outputs: [cells, scores]
```

Note: we are reusing the `application` type for now and use the `application_type` tag to define the workflow, and alternatively we can define a new type `workflow`, but this does mean we need more changes in the website which groups cards by the `type` field. Ideally, we should perhaps show the workflows directly under the `applications` tab. Alternatively, we can also use the tags to define workflows, similar to notebooks.

k-dominik:
[...]

PR #478, opened 2022-10-26
workflow RDF

FynnBe:
implements spec part of #477 
with a few assumptions on open questions...


FynnBe:
A preview of the workflow RDF documentation can be found here:
https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages-workflow_rdf/workflow_spec_0_2.md

FynnBe:
adapted after furrther discussion with @oeway and @constantinpape

example_specs/workflows/hpa/single_cell_classification.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+outputs:
+- name: cells
+  type: tensor
+- name: scores
+  type: tensor
+
+
+steps:
+- op: set_
```
oeway, line 26: `set_outputs`?

example_specs/workflows/stardist/stardist_example.yaml
```diff
@@ -0,0 +1,70 @@
...
+- id: workflow
+  op: run_workflow
+  inputs: [test_tensors.outputs.raw]
+  outputs:
+    - labels
+    - coord
+    - points
+    - prob
+  kwargs:
+    rdf_source: ${{ self.rdf_source }}
```
oeway, line 62: Question: The kwargs can also accept runtime outputs, something like `${{ stepx.outputs.labels }}`, right?

FynnBe: Yes, why not!

example_specs/workflows/hpa/single_cell_classification.yaml
FynnBe:
```suggestion
```

FynnBe: this was just a typo I think..

FynnBe:
hey @oeway !
I was just working on the workflow RDF and found some more aspects we didn't think about before and started discussing quite a bit with @constantinpape .
We really oppose inputs with defaults as it makes it very unclear when to use inputs and when kwargs.

We iterated over several solutions and arrived at this:

- no kwargs field, only inputs
- an input may have a default only if all following inputs have one too
- outputs of one step are always positonal args to the next (like before)
- outputs of one step may be named (like before)
- inputs to a step may be given as a list (like before)
- inputs to a step may be given as a dict (like we gave kwargs)
- if an input dict is missing inputs, these are taken from the args (outputs of the previous step) if present
  - no interleaving: no input following an input given in an input dict can be taken from the present args 

Essentially we don't try to come up with our own logic, but mirror how python callables behave.

FynnBe:
> - inputs to a step may be given as a list (like before)
>  - inputs to a step may be given as a dict (like we gave kwargs)
> ...
> 
> Essentially we don't try to come up with our own logic, but mirror how python callables behave.

been thinking about this part a bit more... 

we can also mix positional inputs and named inputs with list entries containing a dict with a single key (that is the name of an input):
```yaml
steps:
- op: ...
  inputs: 
  - arg0
  - arg1
  - key0: value0
  - key1: value1
```


key word order is ignored:
equivalent to above:
```yaml
steps:
- op: ...
  inputs: 
  - arg0
  - arg1
  - key1: value1
  - key0: value0   # key1 and key2 are given as key word arguments, their order is ignored 👍 
```

Assuming the previous step returns [arg0, arg1], this is equivalent to the above:
```yaml
steps:
- op: ...
  inputs:   # positional inputs (arg0, arg1) are taken from previous step, optional inputs may be overwritten 👍 
    key0: value0
    key1: value1
```

but I would not allow:
```yaml
steps:
- op: ...
  inputs:  # arg0 is filled in from previous step 👎 
  - arg1  
  - key0: value0
  - key1: value1
```
  
and maybe not even specifying optional inputs as a list like:
```yaml
steps:
- op: ...
  inputs:  # arg0 and arg1 are filled in from previous step 👍? 👎?
  - key0: value0
  - key1: value1
```

  
This pattern implies that input values that are dicts with one key need to be given like this:
```yaml
steps:
- op: ...
  inputs:  
    arg0: {key: value}
```
or
```yaml
steps:
- op: ...
  inputs:  
  - arg0: {key: value}
```
because
```yaml
steps:
- op: ...
  inputs:  
  - key: value
```
would be interpreted as the optional input 'key' with value 'value' and not as the first positional argument (here arg0) with value {key: value}.
Of course we could compare the expected arg name and the given key, but I would not rely on that and mandate to specify single key dicts only as a named input.

FynnBe:
and @oeway one more think @constantinpape and I discussed:
We should add `axes` to input if input `type: tensor`.

oeway:
Hi, In the first glance, I like the idea of merging inputs and kwargs. However, when think it a bit further I find it actually making the usage much more complicated and confusing.

> We really oppose inputs with defaults as it makes it very unclear when to use inputs and when kwargs.

I don't really get what this means, the rule is pretty clear to me. Same for defining a python function, if it's required, go for inputs, and if it's optionally with default, go for kwargs. For inputs, you don't need to name them, but kwargs you need to name them. inputs is a list while kwargs is a dictionary. You can of course merge the two by exploiting list of dict, or dict as you are proposing, but feels very hacky and needs a lot of additional rules when using it.

> Essentially we don't try to come up with our own logic, but mirror how python callables behave.

But the actually python callables actually separates the args (a list) and kwargs (a dict). Using a list with 1-key dict or dict to encapsulate the two into one object make it feel hacky and hard to use.

In the following example:
```yaml
  - op: zero_mean_unit_variance
  - op: model_inference
     kwargs:
         model_id: fearless-crab
         preprocessing: false # disable the preprocessing
         postprocessing: false # disable the postprocessing
  - op: stardist_postprocessing
     kwargs:
        diameter: 2.3
```
If I understand correctly, to adapt what your proposed it will become:
```yaml
  - id: step1
     op: zero_mean_unit_variance
     outputs: [out1]
  - id: step2
     op: model_inference
     inputs:
         - step1.outputs.out1
         - model_id: fearless-crab
         - preprocessing: false # disable the preprocessing
         - postprocessing: false # disable the postprocessing
     outputs: [out2]
  - op: stardist_postprocessing
     inputs:
        - step2.outputs.out2
        - diameter: 2.3
```

Note that, because in step2 and step2 we need to set the kwargs, we are forced to name the inputs, and this propagate to all the other steps to force us set id for the first two steps and name the outputs. This introduces a lot of hassle for the naming, and making sure they are identical. 

We can potentially allowing ignoring the positional inputs when we only want to change the kwargs, however, this make the inputs fields looks so confusing, since we are not writing out all the inputs.

Overall, I felt the merging of inputs and kwargs will require more rules, the list of 1-key dict feels hacky, separating them actually make more sense to me. 

> We should add axes to input if input type: tensor.

Sure, maybe we should bring in more tensor spec (e.g. shape) too, however, for simplicity, maybe adding axes, shape, dtype etc. as an optional field would help.

FynnBe:
> Note that, because in step2 and step2 we need to set the kwargs, we are forced to name the inputs, and this propagate to all the other steps to force us set id for the first two steps and name the outputs. This introduces a lot of hassle for the naming.

no. I would not force naming/specifying `outputs` if `inputs` is given.
Also named inputs could be overwritten without touching the positional inputs by providing them as dict:
```yaml
  - id: step0
     op: zero_mean_unit_variance
  - id: step2
     op: model_inference
     inputs:
         model_id: fearless-crab
         preprocessing: false # disable the preprocessing
         postprocessing: false # disable the postprocessing
  - op: stardist_postprocessing
     inputs:
        diameter: 2.3

```

If positional inputs need to be specified this would become
```yaml
  - id: step0
     op: zero_mean_unit_variance
     outputs: [out1]
  - id: somehting-in-between  # added this step to the example, because otherwise out1 would not need naming
     op: ...
  - id: step2
     op: model_inference
     inputs:
         - step1.outputs.out1
         - model_id: fearless-crab
         - preprocessing: false # disable the preprocessing
         - postprocessing: false # disable the postprocessing
  - op: stardist_postprocessing
     inputs:
        diameter: 2.3

```

to refine my idea of merging inputs and kwargs I wanted to propose a slight update:
- if `inputs` are given as a list and the last entry is a dictionary, those are named inputs (kwargs). 
  This avoids a list of len 1 kwargs which indeed feels very hacky.
  
The above example then becomes:
```yaml
  - id: step0
     op: zero_mean_unit_variance
     outputs: [out1]
  - id: somehting-in-between  # added this step to the example, because otherwise out1 would not need naming
     op: ...
  - id: step2
     op: model_inference
     inputs:
         - step1.outputs.out1
         - model_id: fearless-crab
           preprocessing: false # disable the preprocessing
           postprocessing: false # disable the postprocessing
  - op: stardist_postprocessing
     inputs:
        diameter: 2.3

```


**However**, I have to agree with @oeway . this is still hacky...
the issue we tried to address, and the confusion we mentioned before was that inputs may have defaults. So another solution to a clearer specification would be mandatory inputs:
- `inputs` are have no defaults!
- if `inputs`  are not specified previous outputs are given as inputs
- we rename `kwargs`  (too python specific) to `options` (other suggestions welcome!), which always have a default value and a name
- `options` are specified as a dict and always referred to by their name (key), because an arbitrary subset of them may be overwritten. This further distinguishes them from `inputs` and avoids confusing the two sets of parameters.
- [I would like this one, but it is not a must] `inputs` may also be given as a dict via their names, but stay separate from the `options`.


> Sure, maybe we should bring in more tensor spec (e.g. shape) too, however, for simplicity, maybe adding axes, shape, dtype etc. as an optional field would help.

I agree that shape and dtype may be added as optional fields, but I think we cannot handle tensors reasonably without knowing their axes. We don't want a bunch of transpose ops littered throughout the workflows which would be the consequence... 
many of the ops (pre-/ and postprocessing) defined in spec and implemented in core have an `axes` kwarg. (mean per channel etc... )...

FynnBe:
@oeway @constantinpape I know this is a lot to read through and think about, but it would be really convenient for me, if you could take a brief moment and just comment if we can proceed with the current state or need more discussion (if so that does not need to happen right away ;-) )

I tried to comment all discussed/proposed decisions in the following example:
```yaml
inputs:
- name: ipt0
  type: tensor
  axes: bxy      # if type=='tensor' then 'axes' is mandatory

options:   
- name: param0  # although options are given as a dict we specify them as a list to match inputs/outputs pattern
  type: int
  default: null   # options always need a default value (may be null or of type 'type')

outputs: 
- name: ...
  type: tensor
  axes: bxy
- name: ...
  type: bool

steps:
  - id: step0
     op: zero_mean_unit_variance
     outputs: [out1]
  - id: somehting-in-between  # added this step to the example, because otherwise out1 would not need naming
     op: ...
  - id: step2
     op: model_inference
     inputs: [step1.outputs.out1]
# inputs as list or alternatively by name as dict, but either way always all inputs need to be specified:
#     inputs: 
#       raw: step1.outputs.out1
     options:
         model_id: fearless-crab
         preprocessing: false # disable the preprocessing
         postprocessing: false # disable the postprocessing
  - op: stardist_postprocessing
     options:
        diameter: 2.3

test_steps:  # analog to 'steps', but do not take any inputs/options (it may call the 'steps' multiple times with different fixed inputs/options though
  - ...

**Comment by constantinpape** [comment]
@oeway:

> I don't really get what this means, the rule is pretty clear to me. Same for defining a python function, if it's required, go for inputs, and if it's optionally with default, go for kwargs. 

I agree with this. But when I discussed this with @FynnBe, he said that you wanted to allow default values for `inputs`. And this would not make sense to me. Judging from your comment here that might have been a misunderstanding.

> - inputs have no defaults!
>  - if inputs are not specified previous outputs are given as inputs
>  - we rename kwargs (too python specific) to options (other suggestions welcome!), which always have a default value and a name
> - options are specified as a dict and always referred to by their name (key), because an arbitrary subset of them may be > - overwritten. This further distinguishes them from inputs and avoids confusing the two sets of parameters.
> - [I would like this one, but it is not a must] inputs may also be given as a dict via their names, but stay separate from the options.

That sounds good to me. About the last point: I would rather not do it since I think we need to introduce more rules etc. and it makes reading everything more complex. But I don't have a strong opinion on this, so would be fine with this either way.

I also agree that we should have `axes` mandatory for tensors.

> I know this is a lot to read through and think about, but it would be really convenient for me, if you could take a brief moment and just comment if we can proceed with the current state or need more discussion (if so that does not need to happen right away ;-) )

Looks good from my end. My time to look into this will be very limited in the next few weeks, so please don't wait on my feedback to go ahead here.

**Comment by constantinpape** [comment]
> Hi @FynnBe I like this! No strong opinion about changing `kwargs` to `options`. If we do that, would it also make sense to change `list/dict` to `array/object` then? To align with the json types.

I also don't have a strong opinion here, we can also stick to `kwargs` to be consistent with the rest of our spec where we use this.

> A minor suggestion, would it make sense to make `name` field under `inputs` optional? just to align with the ops spec, so we can use the default name `self.inputs.0` (same as unnamed inputs in ops).

I think it would be better to be more explicit here and require the name if we have references to it.

**Comment by FynnBe** [comment]
First of all: Thanks for the fast feedback!

> For axes, would it make sense to bring https://github.com/bioimage-io/spec-bioimage-io/issues/70#issuecomment-825737433 into the tensor spec?

yes, I think it would! I'll get on that... 

> I think it would be better to be more explicit here and require the name if we have references to it.

I agree, let's keep it explicit with mandatory input names. Furthermore, I would like to require the input names to be valid python identifiers for simplicity. '0'--not a valid identifier--would therefore require an exception to that rule... 

> `self.inputs`

this is a minor detail, but if you have no objections I will stay with our previous formulations and reference inputs by `${{ inputs.<input name> }}` (without the `self` scope...

> kwargs/options ... list/dict to array/object 

is `options` a thing in JSON??
actually `kwargs` does not imply that it is optional, therefore I would prefer `options` here. (many of the `kwargs` for pre-/postprocessing in the model spec are mandatory)
I would stick to list/dict as I think it is much more expressive compared to array and object. array is easily confused with tensor and object is very generic.

**Comment by FynnBe** [comment]
from the BioImage.IO meeting: 
named or not named that is the question.
```yaml
  steps:
  - op: my_op
     inputs: [5, 1]  # positional
     inputs:  #  or named
       mean: 5
       std: 1
     inputs:  # the benefit of named inputs -- order does not matter!
       std: 1
       mean: 5

```

```python
# named, mandatory inputs (options would be any kwargs with a default)
def my_op(*, mean, std):
    ...

# positional or named
def my_op(mean, std):
    ...

# if needed: in other langauges we can remember the name order to map named inputs to positional ones
y_op_arg_names = ["mean", "std"]
def my_op(arg1, arg2):
     ...

**Comment by oeway** [comment]
As we discussed, I think it's better to make positional `inputs` actually positional and only support list, not dict. 

In python it directly map to `args`, and it will make it much easier to implement in other language, e.g. javascript too.  In other languages, introducing additional naming (as you proposed) for positional arguments is just another layer of complexity, without that, we will be able to map all the functions in a whole library and make them available as ops.

Even in Python, there are positional-only arguments (see https://peps.python.org/pep-0570/ and you can find the good motivation on why this exists).
```python
def equation(x , y , / , z):
    return x * y + z
#The arguments to the left of '/' (in this case x and y) can only be passed postionally whereas the argument to the right of '/' can be passed positionally or with a keyword 

print(equation(2 , 3 , 5)) #all three arguments are positional 
print(equation(2 , 3 , z = 5)) #the first two arguments are positional while the third is a keyword argument
```
If you do do this, it will raise an error:
```
>>> equation(x=1,y=2, z=3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: equation() got some positional-only arguments passed as keyword arguments: 'x, y'
```

There are functions (e.g. the `print` function) defined as:
```python
def process(*args, **kwargs):
   ...
```
And the named positional arguments won't work in this case. I would see passing positional arguments as keyword arguments is really a very special feature of python, it is not very commonly used too.

Plus when the inputs to the ops are not written, we will go for the positional case anyway, right? If we allow specifying it as a dict, does that mean that a previous op can also return a dictionary and we feed it into the next op? This makes it look very confusing.

Practically, the `mean` and `std` example is not common too, they will likely be `options`, not `inputs`. In most cases, the inputs will be a single tensor, and we won't have tens of inputs in most cases. I am not sure how useful it is to have this additional feature too. 

For the sake of simplicity, I would suggest making positional arguments really positional, names doesn't matter when calling the ops.

oeway:
One more thing, I think it's might be better to change the `inputs` and `outputs` under root as `inputs_spec` and `outputs_spec`, so we are clear this is the type annotation, not the actual inputs as we define for the ops in the steps.

constantinpape:
> As we discussed, I think it's better to make positional `inputs` actually positional and only support list, not dict.

I agree with the only positional arguments for `inputs`, since this will also keep things simpler.
(It also means these should never have default values, but I think we all agree on this now.)

FynnBe:
> For the sake of simplicity...

Yes agreed. Below some points I just couldn't keep to myself, feel free to ignore

> Even in Python, there are positional-only arguments (see https://peps.python.org/pep-0570/ and you can find motivations on why this exists).

just to clarify the situation for Python:
```
def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
      -----------    ----------     ----------
        |             |                  |
        |        Positional or keyword   |
        |                                - Keyword only
         -- Positional only
         
```
(from https://peps.python.org/pep-0570/)
so the question is if we choose `inputs` to be 'positional only', 'positional or keyword', or 'keyword only' 

'positional only' arguments ([the pep @oeway linked](https://peps.python.org/pep-0570/)) were introduced in Python 3.8.
We officially support Python 3.7, so we cannot use them in code (or we require at least py3.8, which might be ok?). Otherwise, from the Python side we only have the option between 'positional or keyword' and 'only keyword' arguments.
('keyword only' were already introduced in Python 3.0; see https://peps.python.org/pep-3102/)


> ```python
> def process(*args, **kwargs):
>    ...
> ```
> And the named positional arguments won't work in this case. I

Even if we only accept key word arguments, one can simply add args as a kwarg (maybe name it 'tensors' or sometheing else more expressive though):
```python
def process(*, args: list, **kwargs):
   ...
```
in the workflow RDF this would look like:
```yaml
steps:
- op: ...
  inputs:
     args: [${{ ... }}, ${{ ... }}]
```
(if we allow both positional and key word arguments this would also not be a problem; we would simply forbid the named inputs for ops using *args, or require them being named 'args' (which would forbid using args as a kwarg name)).

> (It also means these should never have default values,

I agree that we should enforce `inputs` never to have default values, but from the Python side this is not true:
> Therefore, the following would be valid function definitions:
> ```python
> def name(p1, p2, /, p_or_kw, *, kw):
> def name(p1, p2=None, /, p_or_kw=None, *, kw):  # 'only positional' with default
> def name(p1, p2=None, /, *, kw):  # 'only positional' with default
> def name(p1, p2=None, /):  # 'only positional' with default
> def name(p1, p2, /, p_or_kw):
> def name(p1, p2, /):
> ```
(adapted from https://peps.python.org/pep-0570/)

I still think the cleanest, most explicit would be to use 'keyword only' arguments and list all 'keyword only' args with a default as options, and all 'keyword only' without a default as inputs. (pretty much what @k-dominik suggested).
The biggest argument for this is that the order of positional arguments is generally not obvious, thus reading the worklow RDF without directly comparing it to the operator docs is less informative than if the workflow RDF always includes the input names.

In consequence any op should return a dictionary (and not a tuple as we currently assume). Thus output names need to be mapped explicitly as well. 

Naming all op inputs and outputs would certainly make any workflow RDF very readable.

And if we are really that worried about boilerplate code, we can treat missing inputs/outputs as the trivial mapping to shorten cases like this:.
```yaml
steps:
- op: A
  outputs:  # may be left out
     tensor: tensor
- op: B
  inputs:  # may be left out
    tensor: tensor  
```
or
```yaml
steps:
- op: A
  outputs:
     mean_tensor: tensor
- op: B
  inputs:  # may be left out
    tensor: tensor  
```

FynnBe:
> One more thing, I think it's might be better to change the inputs and outputs under root as inputs_spec and outputs_spec, so we are clear this is the type annotation, not the actual inputs as we define for the ops in the steps.

also one more thing, op `outputs` are also not real outputs, but a mapping of outputs to a name. We might want to consider calling the field `output_names` or `map_outputs`... For symmetry I'd be fine with keeping it `outputs`, though.

oeway:
> Yes agreed. Below some points I just couldn't keep to myself, feel free to ignore

Good! Let's go for that!

We actually already discussed the option of doing keyword-only for the workflow during the hackathon, and we decided to go for positional for simplicity if I remember correctly.

For what it worth, I'd like to add some response here:

> 'positional only' arguments ([the pep @oeway linked](https://peps.python.org/pep-0570/)) were introduced in Python 3.8.
We officially support Python 3.7, so we cannot use them in code (or we require at least py3.8, which might be ok?).

This is exactly the reason why we should stick with 'positional only' `inputs`, so we will be all fine with 3.7 and 3.8+ versions. If we do keyword-only argument, it won't work with the positional only functions in python, e.g. `range()` in python 3.8 unless we wrap every such function and make them keyword-only. But, the question is why bother doing that if we have a way to just map them as is, and we might want to expose an entire python module as ops. 

This seems to be a more logical choice to me, to reduce development/maintenance effort, and also simplify the usage for writing workflows as we already discussed.

> The biggest argument for this is that the order of positional arguments is generally not obvious, thus reading the worklow RDF without directly comparing it to the operator docs is less informative than if the workflow RDF always includes the input names.

We need the docs anyway, how would one write the outputs/inputs without reading the docs if we go for keyword-only inputs? I'd argue the contrary, if we go for keyword-only, it make it harder to stack the ops, because we need to carefully read each ops to see what are the keys they accepts and what they returns. One op might return `output` as a key, the other one might return `tensor` or `data`, while it accept `input_tensor` or `x` as inputs. Mapping these is very much error prone, simply because we need to check and write them exactly.

> And if we are really that worried about boilerplate code, we can treat missing inputs/outputs as the trivial mapping to shorten cases like this:.
```
steps:
- op: A
  outputs:  # may be left out
     tensor: tensor
- op: B
  inputs:  # may be left out
    tensor: ${{ A.outputs.tensor }}
```

The issue is that if we do keyword-only, we won't be able to left them out anymore. In the case above, we can only left out the `inputs` if we know exactly the key is called `tensor`, same as the key produced from the last step. Since we won't likely to have the exact key matching between steps, we will have to write it out all the time, and checking docs to retrieve the exact input key for every op, which is a lot more complicated and error prone.

Most importantly, with positional-only inputs and outputs, we simply don't care their name but the order. And the vast majority of the ops only accept 1 tensor in inputs and returns 1 tensor as outputs,  we won't bother writing them out at all -- this greatly simplify the writing of most workflows! 

For most cases, not sure how much it will adds by writing them out as `tensor: tensor`, or `input_tensor: A.outputs.output_tensor`. 

We are designing a new language here, and since most of the other programming languages uses positional arguments for function calls, and they don't see this as the source of error, I don't see why we should force it to keyword-only.

After all, we still have the options, if the developers would like to force keyword only, they can still use the options with default to null, and throw error if the value is null. As a workflow, I don't think we should force everyone to use keyword only.

oeway:
> > One more thing, I think it's might be better to change the inputs and outputs under root as inputs_spec and outputs_spec, so we are clear this is the type annotation, not the actual inputs as we define for the ops in the steps.
> 
> also one more thing, op `outputs` are also not real outputs, but a mapping of outputs to a name. We might want to consider calling the field `output_names` or `map_outputs`... For symmetry I'd be fine with keeping it `outputs`, though.

Good points! I am fine with changing them too, but it looks nicer to use `outputs` as you said.

FynnBe:
> it won't work with the positional only functions in python, e.g. range() in python 3.8 unless we wrap every such function and make them keyword-only

range returns a range object--a generator. I don't think we want support for generators in the workflow definitions, so we'd have to wrap this anyway.

> After all, we still have the options, if the developers would like to force keyword only, they can still use the options with default to null, and throw error if the value is null. 

These implicitly mandatory inputs would lead to runtime errors which could be avoided by statically checking if invalid defaults were not present.

> `skimage.segmentation.watershed`

if you want to support 3rd party functions out of the box, the workflow ops have no chance to be language agnostic. Additionally, if we wanted to use these we would be well advised to wrap them to be more independent of api changes in that 3rd party library, otherwise we will have to pin contrib dependencies very hard.

> We are designing a new language here,

Maybe this is were we go all wrong. Maybe we just shouldn't. If implementations of ops are limited to Python, then maybe the workflows themselves should be treated just as other ops--also defined in Python. A workflow is just another op. Once we agree on how to describe inputs and outputs in terms of function signature and output structure in Python, we could just generate a high level yaml description for non-coders to browse and then focus on the bioimageio.contrib library.
In other words: what if we drop `steps`? Yet another way of thinking about it: Isn't Python the most suitable approach to implement `steps`?
 
I also tried to imagine writing a workflow RDF... The moment it get's complicated I would probably first write a Python script equivalent, which is much easier to debug and then translate to 'workflow RDF', but if all we do is translate the workflow RDF back to Python.... ??

constantinpape:
Sorry, I don't have the time to read through all the discussion now, but from what I understand we actually have a viable solution that we can go with and that everyone is in principle fine with:
`inputs` are positional only and are mandatory (i.e. no default values!) and `kwargs`/`options` (I don't have a strong opinion on what to choose for the name) have default values and are specified by keyword. Leaving the question of how exactly this maps to python, I think this is a good first minimal solution that supports all our initial needs.

> if you want to support 3rd party functions out of the box, the workflow ops have no chance to be language agnostic.
> Additionally, if we wanted to use these we would be well advised to wrap them to be more independent of api changes in that 3rd party library, otherwise we will have to pin contrib dependencies very hard.

I agree that direct support of 3rd party functions is too ambitious in the beginning and will make it not possible to be language agnostic. I think we have to live with the fact that we will need to have a wrapper in these cases.

> I also tried to imagine writing a workflow RDF... The moment it get's complicated I would probably first write a Python script equivalent, which is much easier to debug and then translate to 'workflow RDF', but if all we do is translate the workflow RDF back to Python.... ??

As far as I understand one of the primary goals is to enable running the same workfows in java, and @carlosuc3m has already said he wants to explore implementing these workflows in the bioimageio java libraries.

Overall my suggestion would be:
- implement the minimal viable spec now (see above)
- make a prototype implementation of some contrib ops and workflwos in python and usage in the bioengine
- try porting these to java (@carlosuc3m)
- treat all of this as a prototype without officially declaring it as part of the spec yet, after these first implementations in python **and** java reevaluate the design and see if we need any changes.

carlosuc3m:
Hello everyone,
answering to @constantinpape : 

> As far as I understand one of the primary goals is to enable running the same workfows in java, and @carlosuc3m has already said he wants to explore implementing these workflows in the bioimageio java libraries.

I plan to start implementing this prototype of worflows in the JAva core next week yes. With that I guess I will be able to give some feedback on the robustness of the design. 

Regarding the last point of @constantinpape's list

> try porting these to java (@carlosuc3m)

In principle this should not be necessary to translate the new ops into Java .The idea is to support Python scripting and execution in Java using [JEP](https://github.com/ninia/jep) to avoid code repetition and to maximize reproducibility. A core library that bridges Java and Python through JEP [already exists](https://gitlab.pasteur.fr/bia/jep/-/tree/carlos-edits) and it will be soon integrated in the JAva model runner core.

The library allows automatic installation of miniconda and Python, facilitates installation of Python libraries and enables easy BioImage.io python execution. With this, Java softwares will be able to execute both JAva and Python ops (pre- and post-processing) indistinctly, just parsing what is specified on the rdf.yaml.

What do you think? Would you still rather translate some of the easy ops into JAva? I think that for some cases, such as cell-pose or stadist it does not make sense to have them in Java, but some simple ones might be good to increase the execution efficiency (using Python in JAva implies loading some native libraries).

constantinpape:
Thanks for the clarification @carlosuc3m! I wasn't aware that the JEP integration was this advanced already.

> I think that for some cases, such as cell-pose or stadist it does not make sense to have them in Java,

I agree. (For stardist there actually is a java implementation already that we can use, but indeed porting more complex algorithms is not scalable, so the JEP option is really good!)

> What do you think? Would you still rather translate some of the easy ops into JAva?

I think having examples for one or two simple ops in Java and more complex ones in JEP would be really cool, but it depends on how much extra work that is for you!

constantinpape:
As one more high-level comment: I think I mostly stay by [my comments here](https://github.com/bioimage-io/spec-bioimage-io/pull/478#issuecomment-1301988082). The one thing to consider is that we are re-inventing workflow description languages. BUT the problem with them is that there are many of them, and probably none of them will really fit our use-case. So I think it will take us longer to find a suitable one and adapt it to our purposes compared to implementing the prototype layed out here (if we keep it simple ;)). 

But we should acknwoledge that what we do now is not set in stone and treat this as experimental.

FynnBe:
most content---workflow implementations and workflow exectuion logic--was moved to https://github.com/bioimage-io/workflows-bioimage-io-python
the resolved nodes remain.

PR #481, opened 2022-10-28
Update bio-rdf-validator.imjoy.html

FynnBe:
Should fix: https://github.com/bioimage-io/core-bioimage-io-python/issues/303

### DIFF
--- a/scripts/bio-rdf-validator.imjoy.html
+++ b/scripts/bio-rdf-validator.imjoy.html
@@ -14,7 +14,7 @@
   "api_version": "0.1.8",
   "env": "",
   "permissions": [],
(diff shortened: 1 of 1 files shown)


FynnBe:
changes from 0.4.6 include (from https://github.com/bioimage-io/spec-bioimage-io#changelog):
- Breaking changes that are fully auto-convertible
  - `name` field may not include '/' or '' (conversion removes these)

- Non-breaking changes
  - `cite` field is now optional

- add simple forward compatibility by treating future format versions as latest known (for the respective resource type)
- add `axes` and `eps` to `scale_mean_var`

PR #483, opened 2022-10-31
Add sample model that upscales input.

pattonw:
used for testing core-bioimage-io-python tiling works on models that have a scale factor


constantinpape (approved):
Thanks @pattonw. I will merge this despite the test error (it's unrelated to the changes here).

FynnBe:
was too slow to comment:
would be great to include this model also in a static validation test here in spec... could we add that still?

constantinpape:
> was too slow to comment:
> would be great to include this model also in a static validation test here in spec... could we add that still?

Yeah, makes sense. Can you make a follow up PR?

FynnBe:
will do

FynnBe:
> will do

see https://github.com/bioimage-io/spec-bioimage-io/pull/484

PR #485, opened 2022-11-03
Report on forward compatibility attempt in validation errors

FynnBe:
when treating a future format version as the current latest and errors occur, this attempt at forward compatibility needs to be included in the error report!

This PR does exactly that.


constantinpape (approved):

constantinpape:
The changes look good @FynnBe, but somehow CI isn't running here. Can you check what's wrong with it before we merge it?

FynnBe:
> The changes look good @FynnBe, but somehow CI isn't running here. Can you check what's wrong with it before we merge it?

it runs now on the merge commit with main... not investigating further. maybe we hit some rate limit

PR #486, opened 2022-11-03
Fix processing kwargs

FynnBe:
Found some inconsistencies in the pre- and postprocessing kwargs..
we required `axes` and `mode` kwargs to be specified despite having sensible default values for them. 

Also `scale_linear`'s `gain` and `offset` are conditionally mandatory (at least one needs to be set). The `bioimageio_maybe_required` flag ensures the documentation marks that with an asterisk *.


constantinpape (approved):

constantinpape:
Ok, I will go ahead and merge.

PR #488, opened 2022-11-05
Validate doc field

FynnBe:
as discussed in bioimage-io/core-bioimage-io-python#344


FynnBe:
Here we need to either 
- fix the issue with the `root` context as directly validating a field with a relative path/url is not possible without resolving it or
- move this test to core as a 'dynamic' test when we have the resolved model node

We should probably prefer the latter option as it is easier and does not imply any downloads during static validation which would cause trouble running in pyodide.

PR #489, opened 2022-11-22
elaborate doc for multi-file weights and improve typing with MyPy 0.991

FynnBe:
As discussed in https://github.com/bioimage-io/core-bioimage-io-python/issues/317

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
       if: ${{ matrix.python-version != '3.7' }}
       run: |
         mkdir -p .mypy-cache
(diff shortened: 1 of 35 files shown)


constantinpape (approved):
Changes look good. Any idea why mypy fails now?

FynnBe:
> Changes look good. Any idea why mypy fails now?

Pretty sure it's caused by MyPy Release 0.991...

constantinpape:
Yeah, looks like it, but your fix seems to only work for python 3.7, not 3.8.
I suggest you fix all tests and then just go ahead and merge.

PR #504, opened 2023-03-02
Add tracking of model RDF downloads

FynnBe:


bioimageio/spec/shared/tracking.py
```diff
@@ -0,0 +1,19 @@
...
+    r = requests.get(url)
+    r.raise_for_status()
+
+
+if __name__ == "__main__":
+    model_doi = "10.5281/zenodo.7614645"
+    report_resource_download(model_doi)
+    print(
+        f"https://bioimage.matomo.cloud/?module=API&method=Actions.getDownload&downloadUrl=https://doi.org/{model_doi}&idSite=1&period=day&date=yesterday&format=JSON&token_auth=anonymous"
+    )
```
FynnBe, line 19: I could not yet see the tracking working...

FynnBe: @oeway any idea what I'm doing wrong?

FynnBe:
before merging this PR we should set the "BIOIMAGEIO_COUNT_RDF_DOWNLOADS" env var to "false" in bioimageio.core tests and the collection CI...

oeway:
Hi @FynnBe Thanks for implementing this, please hold on and let me check what's going on with the reporting. I thought it's because the delay (they mentioned the reporting is not in realtime), but it seems not.

bioimageio/spec/shared/tracking.py
oeway: There are two things we need to set here:
 1. we need to configure the `date` properly, it was configured to start from `yesterday` and the period is a `day`, so there won't be any traffic for that model. I would change `period` to `year` and `date` to `2023-03-01` (see here: https://developer.matomo.org/api-reference/Piwik/Period)
 2. the report will only be generated every 15 minutes: https://matomo.org/faq/general/faq_41/ so we won't see the report immediately.
 
 You can see an example here (I triggered the report yesterday so it shows up today): 
 https://bioimage.matomo.cloud/?module=API&method=Actions.getDownload&downloadUrl=https://doi.org/test&idSite=1&idCustomReport=1&period=year&date=2023-03-01&format=JSON&token_auth=anonymous

FynnBe:
ok, looks promising... 
I submitted a report for 'test3':
https://bioimage.matomo.cloud/matomo.php?download=https://doi.org/test3&idsite=1&rec=1&r=646242&h=13&m=35&s=20&url=http://bioimage.io/#/?id=test3&uadata={"brands":[{"brand":"python","version":"3.10"}]}

but also after 15 min I don't get the report back
https://bioimage.matomo.cloud/?module=API&method=Actions.getDownload&downloadUrl=https://doi.org/test3&idSite=1&idCustomReport=1&period=year&date=2023-03-01&format=JSON&token_auth=anonymous

(I immediately got a report for 'test2', so I'm thinking that was you before... ):

FynnBe:
It is hard to test if I successfully disabled tracking in  
- 3631673
- https://github.com/bioimage-io/core-bioimage-io-python/pull/332
- https://github.com/bioimage-io/collection-bioimage-io/pull/552
- https://github.com/ilastik/bioimage-io-resources/pull/68
- bioiomageio.workflows: [d76ea50](https://github.com/bioimage-io/workflows-bioimage-io-python/commit/d76ea50819f84cf89554227591daf92c1d31ff5b)
- Where else are we using bioimageio.spec in tests or CI??

uschmidt83:
FYI, in case you're interested in Plausible.io for analytics. https://github.com/stardist/stardist-napari/pull/12

oeway:
> FYI, in case you're interested in Plausible.io for analytics. [stardist/stardist-napari#12](https://github.com/stardist/stardist-napari/pull/12)

Thanks for the suggestion! I will check it out. With matomo we need to either pay or host our own instance.

@FynnBe The analytics service stopped currently because I didn't pay. Will need to decide whether we run our server, pay, or switch to another option.

uschmidt83:
> With matomo we need to either pay or host our own instance.

Same with Plausible.io.

FynnBe:
@carlosuc3m  this PR adds tracking to bioimageio.spec

PR #508, opened 2023-05-16
add get_resolved_source_path

FynnBe:
for convenient extraction of a source path from any source holding node


k-dominik:
`resolve_source` takes also a list - no?

https://github.com/bioimage-io/spec-bioimage-io/blob/f7e0eaccf59cca4c0cccd294568a61375336e402/bioimageio/spec/shared/_resolve_source.py#L320-L322

but the `resolve_source_path` doesn't. Given the similarity of the name and all, I think it's better to have it explicit.

tests/test_shared/test_shared.py
```diff
@@ -0,0 +1,23 @@
...
+            "https://raw.githubusercontent.com/bioimage-io/spec-bioimage-io/main/example_specs/models/unet2d_nuclei_broad/rdf.yaml"
+        ),
+    ],
+)
+def test_get_resolved_source_path(src):
+    from bioimageio.spec.shared import get_resolved_source_path
+
+    res = get_resolved_source_path(src, root_path=Path(__file__).parent)
+    assert isinstance(res, Path)
+    assert res.exists()
```
k-dominik, line 23: no way to know _what_ path to expect (in terms of where it's pointing to)?

FynnBe: of course for the paths given this is known. 
Also for the url the cache location is known...
but it doesn't really matter.. I don't want to change this test when we change caching.
I'd say if you use this function you don't really care where the path is, but you care that it is a local file path and not a URL

FynnBe:
> `resolve_source` takes also a list - no?
> 
> https://github.com/bioimage-io/spec-bioimage-io/blob/f7e0eaccf59cca4c0cccd294568a61375336e402/bioimageio/spec/shared/_resolve_source.py#L320-L322
> 
> but the `resolve_source_path` doesn't. Given the similarity of the name and all, I think it's better to have it explicit.

ah, I did overlook the list... I type annotated `source` and moved the def below...

tests/test_shared/test_shared.py
k-dominik: One could mock caching though, to make the test independent - in the end you don't want to test caching here...

FynnBe: hmm.. going a bit overboard here I think, but ok...

PR #513, opened 2023-06-05
WIP Improve caching with pooch

### LINKED ISSUE #289: Add proper caching
The current caching solution is appalling.
How about we use https://github.com/reclosedev/requests-cache ?
The default backend seem to be "sqlite", but "filesystem" is also supported, which--I only speculate--might be the better choice for weight files, etc. as we should share the cache across bioimageio.spec and bioimageio.core.
Features also include invalidating cache entries by age.

Obviously we need to add it as non-pyodide dependency (see #265).

@oeway @constantinpape @k-dominik any thoughts/comments?

FynnBe:
WIP, maybe will fix #289 at some point.

### DIFF
--- a/bioimageio/spec/shared/_resolve_source.py
+++ b/bioimageio/spec/shared/_resolve_source.py
@@ -3,6 +3,7 @@ import os
 import pathlib
 import re
 import shutil
+import sys
 import typing
 import warnings
 import zipfile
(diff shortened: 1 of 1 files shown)


FynnBe:
An issue is that pooch depends on requests, which we do not want to depend on in order to run in pyodide. 
I think we can still use pooch, just need to set paths etc manually to have pooch an optional dependency as we now import requests optionally

PR #514, opened 2023-06-06
Added Length validator to the String field class

mese79:
I've added the marshmallow's `Length` validator to the String class in shared fields.  
So, if an empty string is passed for a *required* string field, it will raise the length validation error.
```python
from bioimageio.spec import rdf

author_schema = rdf.schema.Author()
valid = author_schema.validate({
    'name': '',
    'email': 'nobody@gmail.com',
})
print(valid)
# {'name': ['Shorter than minimum length 1.']}
```


FynnBe (changes requested):
We should enforce the string's min_length independent of `required`. 
I would be unexpected to accept an empty string, when the field is annotated with `min_length=1`, only because the string is not required.

mese79:
Ok, I've made a simple change. Please check if this is what you have in mind.

bioimageio/spec/shared/fields.py
```diff
@@ -225,7 +225,15 @@ class Raw(DocumentedField, marshmallow_fields.Raw):
 
 
 class String(DocumentedField, marshmallow_fields.String):
-    pass
+    def __init__(self, *, min_length: int = 1, **super_kwargs):
+        validate = super_kwargs.get("validate", None)
+        if isinstance(validate, typing.Iterable):
```
FynnBe, line 230: I suppose the `validate is None` case is missing...
```suggestion
        if validate is None:
            validate = []
        elif isinstance(validate, typing.Iterable):
```

FynnBe (approved):

FynnBe:
The tests fail, because the URI fields inherites from String and does not implement __len__ ...
I will either fix this later or the changes for the next release will supersede this entirely.

mese79:
I've made an edit: for some the raw_nodes I implemented __len__ methods and for some fields that are subclass of the String I removed the Length validator from the validation list (although this might be not the best solution).

PR #523, opened 2023-09-26
requests: set User-Agent to "ci"  in a CI environment

FynnBe:
helps with https://github.com/bioimage-io/collection-bioimage-io/issues/627


bioimageio/spec/shared/_resolve_source.py
```diff
@@ -466,7 +466,11 @@ def _download_url(uri: raw_nodes.URI, output: typing.Optional[os.PathLike] = Non
             # download with tqdm adapted from:
             # https://github.com/shaypal5/tqdl/blob/189f7fd07f265d29af796bee28e0893e1396d237/tqdl/core.py
             # Streaming, so we can iterate over the response.
-            r = requests.get(str(uri), stream=True)
+            headers = {}
+            if os.environ.get("CI", "false").lower() in ("1", "t", "true", "yes", "y"):
```
oeway, line 470: Can we also support something like `BIOIMAGEIO_USER_AGENT` to override the CI behavior?

PR #529, opened 2023-10-17
Disable url suffix checks

FynnBe:
Due to the recent Zenodo API changes zenodo file urls now terminate with "/content". 
Our current caching mechanism saves them with the filename "content" which leads to issues when validating a files suffix.

Instead of fixing caching I decided to quickly remove all suffix validation as we are close to releasing the major 0.6 version soon (that moves caching entirely to bioimageio.core).

### DIFF
--- a/bioimageio/spec/VERSION
+++ b/bioimageio/spec/VERSION
@@ -1,3 +1,3 @@
 {
-    "version": "0.4.9post2"
+    "version": "0.4.9post3"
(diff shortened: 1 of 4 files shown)


oeway:
Moving caching to core make sense! Thanks!

FynnBe:
thanks to @k-dominik  for finding this bug!

PR #541, opened 2023-11-16
Poetry

ctr26:
Hi again,

Am I ok to add poetry dependency stuff to the pyproject toml?

Thanks

Craig

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,13 +20,15 @@ typing-extensions = "^4.8.0"
 lxml = "^4.9.3"
 torch = "^2.1.1"
 
-[tool.poetry.group.test.dependencies]
-pytest-cov = "^4.1.0"
-black = "^23.11.0"
-pyright = "^1.1.336"
-pooch = "^1.8.0"
-pre-commit = "^3.5.0"
-pytest = "^7.4.3"
+pytest-cov = {version="^4.1.0", optional = true}
+black = {version="^23.11.0", optional = true}
+pyright = {version="^1.1.336", optional = true}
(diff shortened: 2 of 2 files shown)


FynnBe:
I haven't used poetry myself, so please keep that in mind... 
Couple questions
- benefits over mamba?
- dependencies in yet another location? setup.py, environment.yaml... and now poetry? (how) can this be simplified?

stricter pinning which seems to be promoted by the use of poetry seems nice, but we don't want pin too strictly as this library should be easy to add to other envs.

ctr26:
Poetry is mamba compatible, poetry just completely replaces the old setup.py stuff. 

So with this implementation you can run `pip install .` or  `pip install pip install git+https://github.com/bioimage-io/spec-bioimage-io.git` or people who dont want to use conda for instance

ctr26:
Realistically your env file could just then be 

```

channels:
- anaconda
- defaults
dependencies:
- pip
- pip:
  - -e .
```

ctr26:
Poetry also has built in functionality for pushing and building straight to Pypy

FynnBe:
ah, so we can remove setup.py then?
but it's not possible to install packages from conda channels using poetry?

ctr26:
Exactly.

ctr26:
As for the current error on ci, pip is supposed to work out of the box with pyproject tomls that have poetry in them, so I need to figure out why it's failing

ctr26:
More generally I tend to only use conda for installing packages that don't work outright on pypi, like the cudatoolkit; which is usually compiled binaries I've found.

jmetz:
Ahh, the good old dependency management fun-and-games 😅

I haven't used poetry extensively but my initial impressions are that it's reasonably good, and adding the option of using poetry seems like a good idea. 

On a related note, I'm keeping an eye on rye (rye-up.com), as it seems like it has a lot of potential to solve a lot of the python installation/distribution/dependency management issues. That said, it _could_ end up being just another anaconda/mamba etc etc... Though Armin's contributions to the python ecosystem have been pretty solid in the past, so I'm hopeful about this latest one.

FynnBe:
still not 100% onboard, but doesn't look like it hurts, if you maintain it.
It shouldn't interfer with the cond build though: `ModuleNotFoundError: No module named 'poetry'` ... so if it could be separate. like an optional dependency that would be great.. or is that not feasible somehow??

FynnBe:
closing due to inactivity

PR #567, opened 2024-03-18
0.5.1

### LINKED ISSUE #462: Slash in the name
It looks like the validator in the uploader doesn't throw error when there is a slash in the model name.

FynnBe:
add 'ensure_dtype' operations to default pre- and postprocessing
and other changes, see changelog in README.md

Fixes #462


bioimageio/spec/model/v0_5.py
```diff
@@ -1103,7 +1115,14 @@ class InputTensorDescr(TensorDescrBase[InputAxis]):
...
 
     preprocessing: List[PreprocessingDescr] = Field(default_factory=list)
-    """Description of how this input should be preprocessed."""
+    """Description of how this input should be preprocessed.
+
+    notes:
+    - If preprocessing does not start with an 'ensure_dtype' entry, it is added
+      to ensure a tensor's input data type matches the tensor's data description.
+    - If preprocessing does not end with an 'ensure_dtype' or 'binarize' entry, it is added
+      to ensure preprocessing steps are not unintentionally chaning the data type.
```
FynnBe, line 1124:
```suggestion
    notes:
    - If preprocessing does not start with an 'ensure_dtype' entry, it is added
      to ensure an input tensor's data type matches the input tensor's data description.
    - If preprocessing does not end with an 'ensure_dtype' or 'binarize' entry, an 
      'ensure_dtype' step is added to ensure preprocessing steps are not unintentionally 
      changing the data type.
```

PR #584, opened 2024-04-15
add input axis flag `concatenable`

FynnBe:
initiated by @constantinpape feedback to the current version and as discussed with @constantinpape and @Tomaz-Vieira separately offline: we need to distinguish between the sample axis length and the block axis length, i.e. we need to know if the axis described has the size of the block or the sample.


FynnBe:
merged for now. it's a minor change and purely optional fields, we can improve on this in future releases if needed

PR #593, opened 2024-05-07
Example notebooks and documentation

esgomezm:


esgomezm:
@FynnBe  here is the new notebook and the doc updated but all the test to the PR fail and I really don't know why

FynnBe:
> @FynnBe here is the new notebook and the doc updated but all the test to the PR fail and I really don't know why

test failiure was unrelated and hopefully fixed now

.DS_Store
FynnBe, line 1: this file shouldn't be here?

FynnBe (changes requested):
couldn't review in detail now... just one minor file too many ;-)

PR #609, opened 2024-06-13
use attribute docstrings

FynnBe:
cc @esgomezm 

this will add all field docstrings to the json schema


esgomezm:
awesome! It means that there will be a short description for each item?

FynnBe:
yes, for most. (and for those few missing we should add it in the Python code ;-) )

PR #613, opened 2024-06-13
Automatic documentation

esgomezm:
Automatic documentation generation using [sphinx for pydantic](https://sphinx-pydantic.readthedocs.io/en/latest/)

- All the files (`conf.py` and `.rst`) are inside the `/scripts/sphinx/source` folder.
- After the installation of requirements, run `make clean` followed by `make html` to compile the HTML with the documentation. 
- The new HTML doc is placed in `/scripts/sphinx/build` 

@Tomaz-Vieira @FynnBe and @oeway I started this but I'm not sure if the result is good, and if it can get any better using sphinx. Maybe heritageing some pydantic classes? The doc still looks quite complex to me.

PR #614, opened 2024-06-13
POC of using sphinx + pydantic

Tomaz-Vieira:
This demonstrates how we could use sphinx to do documentation, embedding the docstring-based documentation for the pydantic models into the more human-ffriendly, manually written text

image

### DIFF
--- a/scripts/generate_sphix_docs.py
+++ b/scripts/generate_sphix_docs.py
@@ -1,0 +1,13 @@
+import subprocess;
+from pathlib import Path
+import os
+
+try:
+    import sphinxcontrib.autodoc_pydantic # type: ignore
+except ImportError as e:
+    import sys
+    print("You need to install 'autodoc_pydantic'. Try running \npip install autodoc_pydantic", file=sys.stdout)
+    exit(1)
+
+os.chdir(Path(__file__).parent.parent)
+_ = subprocess.run(["sphinx-build", "-M", "html", "sphinx_docs/source", "sphinx_docs/build/"])
--- a/sphinx_docs/Makefile
+++ b/sphinx_docs/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = source
+BUILDDIR      = build
+
+# Put it first so that "make" without argument is like "make help".
+help:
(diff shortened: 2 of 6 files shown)


FynnBe:
great stuff @Tomaz-Vieira , Thanks!
@esgomezm  we should explore this a bit further, don't you think?

esgomezm:
Hi! It's very similar to what I added in my [PR](https://github.com/bioimage-io/spec-bioimage-io/pull/613). As said before, I think we should find a way of making this much simpler, organised and human-readable. If you know how to do that with sphinx, I'm happy to help. I couldn't get much further than this

Screenshot 2024-06-13 at 15 27 34

Screenshot 2024-06-13 at 15 27 42

Screenshot 2024-06-13 at 15 28 30

Tomaz-Vieira:
That looks great, @esgomezm  ! Whatever I tried on my end it seemed like it would always leave out some fields, but yours looks pretty complete =) I guess even too complete, since we can see also the `pydantic`-specific fields like `model_config` and `model_fields`. I know next to nothing about sphinx (and specially about this pydantic plugin) so I don't have more insights here. 

It's nice that we can produce these docs, but I'm skeptical as to how much it'll solve our issue; The user-facing, tutorial-like strings would still have to live somewhere next to the docstrings, but such friendly instructions might feel out of place in the midst of the source code. Also, if we write that documentation somewhere else, then it makes it easier for it to go out of sync with the spec itself.

My interest with this is that I would like to put very polite tooltips in the model builder GUI to orient users into building their models, and in a perfect world I'd be able to fetch some of that text from this endeavor. That might require tagging the text somehow so that it can be later extracted and potentially ignored when creating the more blunt developer documentation

esgomezm:
Hi @Tomaz-Vieira , I also have no experience with sphinx. It was a quick try =)

In any case, I fully agree with your. This still needs further plain text description for either a GUI or human readable documentation. Unless someone knows very well how to edit sphinx and its descriptions, I would rather generate a json file.
In this json, we could update the "description" added within the docstring in the code when generating it. 

For now, by the way, I have some code that creates a drafted json like this:

Screenshot 2024-06-14 at 13 15 49
Screenshot 2024-06-14 at 13 15 19

esgomezm:
I basically parsed the json schema generated by the pydantic code from the core library. It's the one in gh-pages branch. I can make a PR. Basically it's a loop through the dictionary keys for which we can add any sort of conditions and text :)

Tomaz-Vieira:
I wonder if we couldn't attach this information directly to the pydantic models, so that we skip one (potentially lossy) transformation to the Json schema, and so that it's harder for the extra text to go out of sync with the spec

FynnBe:
closing in favor of https://github.com/bioimage-io/spec-bioimage-io/pull/630 and https://github.com/bioimage-io/spec-bioimage-io/pull/617

PR #625, opened 2024-08-28
Scripts for generating interactive spec docs

Tomaz-Vieira:
Adds scripts for traversing the spec types and building HTML documentation

The code for generating the docs still relies on some python `3.12` typing features, so it won't pass pyright checking since it's configured in pyproject.toml to use python `3.8`. This can be fixed soon, though

Usage, for now:

```shell
cd ./scripts
PYTHONPATH=.. python -m docgen
```

A million warnings get emmited and the docs are generated at `dist/interactive_docs*.html`. The warnings are usually stuff we could fix in the spec itself (like missing examples)


Tomaz-Vieira:
Closed in favor of #630, which is inside this repo and therefore can pass the deployment test

Tomaz-Vieira:
Closed in favor of #630, which is in this repo and therefore can pass the deployment test

PR #627, opened 2024-09-02
Fixes breaking pyright checks across py versions

Tomaz-Vieira:
Pyright was failing depending on the python version being used; It seems the type hints available change depending on the python version, either because the hints are gated behind version checks or because the dependencies change versions to stay compatible with the python version.

This patch squelches some warnings and tells pyright to use the version of the current interpreter instead of always using `3.8` from `pyproject.toml`


Tomaz-Vieira:
I think one of the tests is failing because the PR comes form a fork. Clsing and re-opening with a branch in this repo.

PR #630, opened 2024-09-05
Scripts for generating interactive spec docs

Tomaz-Vieira:
Adds scripts to generate flattened, interactive, html documentation.

image

Scripts will traverse the python typing hierarchy and produce nested tables with yaml examples at every level of the hierarchy. This gives us maximum flexibility both for rendering as well as for interacting with the documentaiton.

The logic can then be further expanded to add more metadata, like documentation for GUIs, that can be picked up by other external tools


README.md
```diff
@@ -44,7 +44,7 @@ Simplified descriptions are available as [JSON schema](https://json-schema.org/)
 | bioimageio.spec version | JSON schema |
 | --- | --- |
 | latest | [bioimageio_schema_latest.json](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/bioimageio_schema_latest.json) |
-| 0.5 | [bioimageio_schema_v0-5.json](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/bioimageio_schema_v0-5.json) |
+| 0.5 | [bioimageio_schema_v0-5.json](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/bioimageio_schema_v0-5.json) [rendered](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/interactive_docs_v0-5.html) |
```
FynnBe, line 47: the rendered dev docs are more than just documenting the json schema, as they traverse the Python source code, right?
should we link them separately then?

FynnBe (approved):
Didn't go through the actual script, but as that is somewhat independent from the rest of the code base I'd say we can go ahead an merge this.
(We can change the readme linking these new docs afterwards as well)

PR #677, opened 2024-12-19
remove 'other'

FynnBe:


FynnBe:
after some discussion with @akreshuk I removed some redundant categories.
We still need to:
 - [ ] add tags to categories 
     - [ ] that are in use on bioimage.io today
         - [ ] map to a tag included in the EDAM ontology
     - [ ] that are specified by EDAM? (maybe not for now to not clutter the suggestions with unused tags!)
 - [ ] add categories and tags for non-model descriptions (at least for "notebook" and "dataset")
 - [ ] add more aliases to tag_aliases.json (to help informing model creators choosing the same tags)
 - [ ] add pytest tests for tag consistency checks
     - no duplicates across categories
     - aliases exist and not duplicated
 - [ ] consider using yaml files (to use tags for repeating tag categories across types)
     - [ ] generate the static json files from that yaml source in this repo?

PR #685, opened 2025-02-03
Fix opt fields in docs

Tomaz-Vieira:
Also checks for `default_factory` in fields to determine whether they are optional or not


FynnBe (approved):
LGTM

PR #686, opened 2025-02-05
Reduces border and padding noise

Tomaz-Vieira:
The interactive docs were relying a bit too much on borders and padding that created a bunch of visual noise. This PR eases that a bit while still maintaining a clear division between the fields:

Old (redundant borders and padding to the right):

image

New (removed borders and right-padding, trust indenting to represent nesting):

image


FynnBe (approved):
Great stuff!

PR #710, opened 2025-07-29
FAIR fields and Hypha Upload

FynnBe:


FynnBe:
A field that not needed for executing/using a model/dataset/notebook, but that is part of the metadata making the resource FAIR.
(Just because that's probably where you have your example from; the `FAIR[int]` annotation in the test is pretty meaningless).
`my_field: FAIR[Optional[int]] = None` would make `my_field` optional, but if set to None (or an empty sequence, e.g. `FAIR[str] = ""`) an alert (warning with slightly higher warning level than a regular warning) is raised that this field is needed to make the description FAIR.

This will make it easier to tinker with models etc. and allow core to work with models that do not fully comply with the FAIR part of the metadata, e.g. not failing prediction because authors are incorrectly specified.

PR #752, opened 2026-04-18
feat: custom postprocessing ops (inline + built-in, resolves #751)

### LINKED ISSUE #751: Postprocessing vocabulary cannot express flow-dynamics / NMS decoders used by modern instance-segmentation models
## Summary

The current postprocessing vocabulary in v0.5 (`sigmoid`, `softmax`, `binarize`, `clip`, `scale_linear`, `scale_range`, `ensure_dtype`, …) can express the tail of a semantic-segmentation model but cannot express the tail of modern **instance-segmentation** models, where the network produces a dense auxiliary field that has to be decoded into per-object labels by an iterative / non-trivial algorithm that lives outside the neural network.

Concrete examples we'd like to publish on bioimage.io:

| model | network output | postprocessing needed |
| --- | --- | --- |
| **Cellpose / Cellpose-SAM** | `[flow_y, flow_x, cellprob]` | `cellpose.dynamics.compute_masks` — gradient tracking + connected components |
| **Omnipose** | euclidean flow + boundary | distance-transform-based watershed + Euler integration |
| **StarDist** | object probabilities + ray distances | NMS on rays → polygon/polyhedron rasterisation |
| **HoVer-Net** | horizontal + vertical gradient maps | marker-controlled watershed (custom) |

Today the only way to ship such a model on the zoo is the workaround I used for `Cellpose-SAM`: declare the *raw field* as the model output and tell users in the README to run the postprocessing step themselves with the vendor's Python package. That works, but:

1. The model card looks like a semantic-segmentation model, which is misleading; curators and users see a 3-channel logit tensor and have no standard way to turn it into a label map.
2. Zoo consumers (Ilastik, ImageJ, deepImageJ, the bioengine test runner) cannot execute the model end-to-end — they have no way to discover that a vendor-specific Python postprocessing step is required.
3. Every integration rewrites the same algorithm (there are at least three separate implementations of `compute_masks` in the wild already).

## Proposal — `custom` postprocessing op

Add a single new postprocessing descriptor modelled on `weights.pytorch_state_dict.architecture`:

```yaml
outputs:
  - id: masks
    axes: [...]
    postprocessing:
      - id: custom
        kwargs:
          source: flow_dynamics.py      # bundled in the package
          callable: compute_masks       # name of the Python callable
          sha256: ...                   # reproducibility hash
          environment: environment.yaml # reuse weights.*.dependencies semantics
          kwargs:                       # passed through to the callable
            cellprob_threshold: 0.0
```

Semantics:

- The callable receives the output tensor (`xarray.DataArray` or `np.ndarray`, matching the existing `*ProcessingDescr` conventions), returns a tensor with the declared output axes.
- Runs in an isolated conda/venv environment driven by the same `environment.yaml` / `dependencies` mechanism that `architecture.dependencies` already uses — no change to the JSON schema of dependencies, just reuse.
- Consumers that don't want to spawn a Python process (e.g. pure-Java deepImageJ) can still introspect the output tensor shape and flag the model as "needs vendor postprocessing".
- Marked clearly in the spec as a trust boundary — executing arbitrary vendor code. Curator review already gates the zoo, so this is acceptable; the alternative (everyone rolling their own sidecar registry) is worse.

This mirrors the shape of `architecture.source` / `architecture.callable`, which the community is already comfortable with.

## Alternative — a registry of named postprocessing algorithms

If executing arbitrary Python is too large a trust expansion, consider a registry of well-known non-trivial postprocessing ops:

```yaml
postprocessing:
  - id: cellpose_flow_dynamics
    kwargs:
      cellprob_threshold: 0.0
      flow_threshold: 0.4
```

…where the spec explicitly lists `cellpose_flow_dynamics`, `stardist_nms`, `hovernet_watershed`, etc., and each consumer SDK is expected to ship an implementation. This keeps the spec closed but grows linearly with the ecosystem.

## What I'd like feedback on

- Do the maintainers have a preferred direction between the two options above (or a third)?
- Is there already a WIP RFC or design note for this that I missed?
- Happy to prototype the `custom` descriptor against `bioimageio.spec` + `bioimageio.core` and open a PR if the direction is agreed.

## Context

Hit this while preparing a bioimage.io submission for Cellpose-SAM (the current Cellpose 4.x SAM-based model). The submission is now staged as a raw-flow model, but losing the postprocessing step in the spec feels like a gap worth fixing before more instance-seg models land.

oeway:
## Summary

Implements custom pre/postprocessing support for bioimage.io models, addressing the need raised in #751.

### Design

A new `CustomPostprocessingDescr` (id: `"custom"`) is added to the `PostprocessingDescr` union. It provides two usage modes:

**1 — Inline (development phase):** Ship the op as a `.py` file alongside model weights.

```yaml
postprocessing:
  - id: custom
    callable: my_postprocess
    source: my_postprocess.py
    sha256: <sha256 of the file>
    kwargs:
      threshold: 0.5
```

**2 — Built-in (after promotion to `custom_ops/`):** Drop `source` + `sha256`; everything else is identical.

```yaml
postprocessing:
  - id: custom
    callable: my_postprocess
    kwargs:
      threshold: 0.5
```

### Two supported op styles

**Style 1 — Callable class:**
```python
class my_postprocess:
    def __init__(self, threshold=0.5):
        self.threshold = threshold
    def __call__(self, *arrays):
        return (arrays[0] > self.threshold).astype(np.uint8)
```

**Style 2 — Factory function:**
```python
def my_postprocess(threshold=0.5):
    def run(*arrays):
        return (arrays[0] > threshold).astype(np.uint8)
    return run
```

Runtime behaviour is identical: `op = callable(**kwargs)` once; `result = op(*tensors)` per image.

### What's new

- `CustomPostprocessingDescr` in `src/bioimageio/spec/model/v0_5.py`
- `custom_ops/` directory with:
  - `cellpose_flow_dynamics.py` — first built-in op (Cellpose/Cellpose-SAM flow decoding)
  - `README.md` — contributor guide: inline→built-in workflow, both styles, rules
- 6 new tests in `tests/test_model/test_v0_5.py`

### Validation

- `sha256` required when `source` given (enforced via `model_validator`)
- All 6 new tests pass

cc @FynnBe @akreshuk — continuing the discussion from #751. This design keeps the API minimal (one flat descriptor, four fields) while enabling an inline-to-shared promotion path with zero changes to `callable`, `kwargs`, or runtime behaviour.


FynnBe:
I commented in https://github.com/bioimage-io/spec-bioimage-io/issues/751#issuecomment-4286942690

FynnBe (changes requested):
- op implementation should go into bioimageio.core.
- `CustomPostprocessingDescr` should inherite from `FileDescr` for sha256 validation.
- source and sha256 should not be optional at all, no?

oeway:
@FynnBe CI is green on this PR — ready for review and merge.

Summary of changes:
- `CustomPostprocessingDescr` now inherits from `FileDescr` (sha256 validation built-in)
- `source` is required (not optional); `sha256` enforced via `model_validator`
- `_serialize` uses `package_file_descr_serializer` for correct model packaging
- JSON schema updated with `CustomPostprocessingDescr` added to the postprocessing `oneOf`
- Op implementations go in `bioimageio.core` (see companion PR bioimage-io/core-bioimage-io-python#489)

FynnBe:
Please comment on https://github.com/bioimage-io/spec-bioimage-io/issues/751#issuecomment-4286942690

oeway:
Hi @FynnBe thanks for working on this! Please make changes as you like, I would like to move faster with this.

oeway:
@FynnBe are we ready to merge?

PR #768, opened 2026-08-25
Add MicroSamWatershedDescr postprocessing

nilsmechtel:
## Motivation

micro-sam models with the additional instance segmentation (AIS) decoder predict three dense maps — foreground probability, center distance and inverted boundary distance. Turning those maps into instance labels requires a seeded watershed, which today only exists inside the `micro_sam` package. Making it a named postprocessing op lets a packaged, prompt-free micro-sam AIS model declare the step in its RDF instead of shipping a custom source file.

This adds the descriptor only. The implementation lives in bioimageio.core (bioimage-io/core-bioimage-io-python#501), whose operator needs a `bioimageio.spec.model.v0_5` type to bind `from_proc_descr` to.

## What is added

- `MicroSamWatershedKwargs` and `MicroSamWatershedDescr` (`id: microsam_watershed`), following the `CellposeFlowDynamicsDescr` precedent.
- Registration in the `PostprocessingDescr` union, and in the postprocessing-only exclusion tuple of `_InputTensorConv` (a postprocessing op cannot appear in a converted v0_4 preprocessing list).
- Regenerated `bioimageio_schema.json`.

Defaults match `micro_sam.instance_segmentation.InstanceSegmentationWithDecoder.generate` / `torch_em.util.segmentation.watershed_from_center_and_boundary_distances`: thresholds 0.5, foreground smoothing 1.0, distance smoothing 1.6, `min_size` 0. `output_dtype` mirrors `CellposeFlowDynamicsKwargs`. All fields have defaults, so `kwargs` may be omitted entirely.

## Format version

Not bumped here, mirroring the commit that introduced `CellposeFlowDynamicsDescr` (d53a39e8), which landed while `implemented_format_version` was still the previously released `0.5.9` and was bumped separately later. Happy to fold a bump to model `0.5.15` into this PR if you would rather release it that way.

## Test plan

- `MicroSamWatershedDescr` validates through the `PostprocessingDescr` discriminated union, with and without explicit kwargs.
- JSON schema regenerated with `scripts/generate_json_schema.py`; the diff is additive only.

### DIFF
--- a/src/bioimageio/spec/model/v0_5.py
+++ b/src/bioimageio/spec/model/v0_5.py
@@ -1374,6 +1374,54 @@ class CellposeFlowDynamicsDescr(NodeWithExplicitlySetFields):
     kwargs: CellposeFlowDynamicsKwargs
 
 
+class MicroSamWatershedKwargs(KwargsNode):
+    """key word arguments for [MicroSamWatershedDescr][]"""
+
+    center_distance_threshold: float = 0.5
+    """Center distance predictions below this value are used to find seeds."""
+
+    boundary_distance_threshold: float = 0.5
+    """Boundary distance predictions below this value are used to find seeds."""
+
+    foreground_threshold: float = 0.5
+    """Foreground predictions above this value make up the foreground mask."""
+
+    foreground_smoothing: float = 1.0
+    """Sigma of the gaussian smoothing applied to the foreground prediction to
+    avoid checkerboard artifacts. Set to 0 to disable smoothing."""
+
+    distance_smoothing: float = 1.6
+    """Sigma of the gaussian smoothing applied to both distance predictions.
+    Set to 0 to disable smoothing."""
+
+    min_size: int = 0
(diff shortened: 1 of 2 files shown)


FynnBe:
Would it maybe make sense to split this micro-sam postprocessing into generic bits 

1. binarize (distance maps)
2. (seeds from) connected components
3. seeded watershed with foreground mask
4. object size filter.

This might be a chance to add connected components and seeded watershed individually, which are lower in complexity and which might be reused by other postprocessing workflows.

Side note: With sufficient halo binarize and seeds from connected components could run on tiles, the intermediate output (seeds + foreground mask) would be very compressible, great for remote processing! The watershed could be run on the full (stitched) sample (possibly locally in case of remote inference). Point being that splitting this postprocessing into individual parts might benefit deployment, too. 

Or is there added secret sauce in what micro-sam does? 
Maybe @constantinpape wants to chime in on a design decision here?

edit: added side note

Issue #670, opened 2024-11-29
Add parameters input for non-tensor model inputs


FynnBe:
The most recent time this came up is https://github.com/bioimage-io/core-bioimage-io-python/issues/439

Some models may accept additional parameters **after** initialization as part of the forward method. (Not all model key word arguments would be suitable as sample parameters!)

What I have in mind when thinking about this is roughly the following
```yaml
inputs:
  - id: tensor_as_ususal
    description: `An input tensor as the spec allows today already`
    axes: ...
  - parameters:
      - name: diameter
        description: "expected diameter in pixels of objects to segment"
        default: 30     
```

Would be pretty straight forward to add to spec and core, but we should collect some more actual use-cases to discuss if (and how) we want to add this and how challenging it would be to support this in other partner software.
One approach to make adaption easier would be that all parameters have to be optional and some software just doesn't allow users to change those default values... 

cc @bioimage-io/spec-dev

mese79:
If those parameters are related to the model itself, it's better to have them in the `model` section rather than `inputs`.  
Also, some models accept arguments for initialization, and setting them after couldn't be effective. Right now, we have them as `kwargs` in the model section, but I'm wondering if it is possible to `init` the model after `loading` via core.
```python
model = load_description(...)
model.init(a=1, b=2, c="3")
```

FynnBe:
those lines are really doing different things. here is some pseudo code to clear things up:

```python
model_descr = load_description()  # spec
assert isinstance(model_descr, ModelDescr)

# create adapter that initializes the model
adapter = create_model_adapter(model_descr)  # core  

adapter.forward(...)

nn_module = PytorchAdapter.get_network(model_descr.weights.pytorch_state_dict)  # or get the nn_module directly (only for pytorch state dict..)
```

yashraj4:
Hey @FynnBe @mese79 , I went ahead and opened a PR for this (#772). Figured it'd be easier to discuss with actual code rather than just describing it.

To address the points raised here:

On whether these should be under `inputs` or `model` - I put them under `inputs` because these are forward-method arguments (like diameter in Cellpose), not model initialization config. The model section already has kwargs for initialization. But I could see the argument either way, so happy to move them if the maintainers think model is a better fit.

[...]

PR #772, opened 2026-09-14
Add parameters input for non-tensor model inputs (#670)

yashraj4:
## Summary

Adds support for non-tensor model parameters as part of the input spec, addressing [#670](https://github.com/bioimage-io/spec-bioimage-io/issues/670).

Some models accept additional parameters after initialization as part of the forward method (e.g., expected diameter in pixels for segmentation models). This PR adds an optional parameters field to InputTensorDescr so that these can be declared in the spec.

## Changes

- **ParameterDescr** type with fields:
  - 
ame: parameter name passed to the forward method
  - description: short description (max 128 chars)
  - default: default value (required, so all parameters are optional for consumers)
  - dtype: optional data type hint (loat32, int, ool, etc.)
- **parameters** field added to InputTensorDescr in both v0_4 and v0_5 specs
- Tests for ParameterDescr and InputTensorDescr with parameters
- Changelog updated

## Example usage

\\\yaml
inputs:
  - id: image
    axes:
      - type: batch
      - type: channel
      - type: space
        id: x
      - type: space
        id: y
    parameters:
      - name: diameter
        description: expected diameter in pixels of objects to segment
        default: 30
      - name: threshold
        description: segmentation threshold
        default: 0.5
        dtype: float32
\\\

## Notes

- All parameters have defaults so consumer software can safely ignore them
- This is a non-breaking, backward-compatible change
- No format version bump needed since new fields are optional


FynnBe:
Hi @yashraj4 great initiative with this PR! Thanks!

Not sure if we really should go forward with this right now, see https://github.com/bioimage-io/spec-bioimage-io/issues/670#issuecomment-5678418190

Couple small thoughts now that I looked into it already:
- We should not add this to model 0.4, all new models should follow 0.5 anyway...
- I suggest renaming/replacing/removing `dtype` to avoid confusing with an array data type.

yashraj4:
Good point  - `dtype` is overloaded in scientific Python and could cause confusion.

Options:
1. **Remove it entirely**  - the default value already implies the type (`30` → int, `0.5` → float), so it's arguably redundant
2. **Rename to `type`** - but this conflicts with other fields in the spec
3. **Rename to `value_type`** -  clearer that it describes the parameter's value type, not an array dtype

I'd lean toward removing it. It's the simplest option and the default already carries the type info. If someone needs to enforce `"30"` as a string instead of `30` as an int, that's an edge case we can handle later.

What do you think?

FynnBe:
I'd lean towards removing it as well. 
Depending on arising use-cases we might want to have a 'choice' field for valid values or a range with min/max values. But I would like to wait with this feature for actual use-cases to emerge.

